PR #2, opened 2018-03-15
Add ability to perform basic volume actions.

displague:
I was just taking a peak at this repo and figured I'd try adding something to get a feel for it.

I don't know how to get the test/vcr data captured, but I'm sure you will need that updated for volumes.

I think you may have some things in the fixtures related to custom stack scripts that you may not want in this repo publicly (certs maybe?).

I think it may end up being strange [to have some functions take an "options" arguments](https://github.com/chiefy/go-linode/pull/2/files#diff-8c844552c440c1a179655f8ba2f81d11R78) (to only define two options) when other functions just accept each scalar value as an argument.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,7 @@ This project is not ready for production usage, use at your own risk.
 - `/linode/instances/$id/shutdown`
   - [x] `POST`
 - `/linode/instances/$id/volumes`
-  - [ ] `GET`
+  - [X] `GET`
 
 ### Backups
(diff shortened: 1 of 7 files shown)


chiefy:
thanks @displague i'll try to take a look at this asap.

chiefy:
@displague merged in your PRs, thanks for the contribs! I am hoping to work more on this in the next few weeks.

Issue #5, opened 2018-03-19
Implement / expose ability to page through responses


chiefy:
Check https://github.com/digitalocean/godo for their implementation

chiefy:
Check https://github.com/digitalocean/godo for their implementation

PR #7, opened 2018-04-21
initial pass at pagination (Kernels, Linodes)

displague:
A workable approach to pagination (#5).  (I also included an initial CreateInstance, but can move that to another PR)

There is room to simplify this through interfaces.  This PR is mostly intended for discussion.

If a `ListSomething` endpoint is given a `nil` `ListOption` it will automatically paginate and return all results (error handling is still needed for the subsequent requests).

If a `ListSomething` endpoint is given a `ListOption`, it will fetch the requested page and populate the `Results` response from the server.

You can see that this is finding all of the Linode Kernels:
```
$ ./example  2>&1 | grep version | wc -l
     218
```

### DIFF
--- a/API_SUPPORT.md
+++ b/API_SUPPORT.md
@@ -4,7 +4,7 @@
 
 - `/linode/instances`
   - [x] `GET`
-  - [ ] `POST`
+  - [X] `POST`
 - `/linode/instances/$id`
   - [x] `GET`
   - [ ] `PUT`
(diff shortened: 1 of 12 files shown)


displague:
I've attempted to abstract this a bit, https://github.com/chiefy/go-linode/commit/792d5c93ed88a338aac2a883e76c4d30bf979696, but I'm not getting the interface use correct.  When I attempt to use `l *PagedResponse` it is always `nil` (see the comments).

PR #9, opened 2018-04-29
My Stackscripts

dvestal:
It is much more helpful for me to be able to list only *my* stackscripts rather than all public ones. In general most "list" commands are dealing with only your resources so I created the method ListAllStackScripts and changed that of ListStackScripts to include the "mine":true filter.

I do not mind to provide a different implementation of this if you'd prefer.

### DIFF
--- a/stackscripts.go
+++ b/stackscripts.go
@@ -40,15 +40,15 @@ func (l *LinodeStackscript) fixDates() *LinodeStackscript {
 
 // ListAllStackscripts gets all public stackscripts
 func (c *Client) ListAllStackscripts() ([]*LinodeStackscript, error) {
-	return c.listStackScripts(false)
+	return listStackScripts(c, false)
 }
 
 // ListStackscripts gets all my stackscripts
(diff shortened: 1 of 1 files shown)


displague:
@dvestal I have some work towards generalizing (templatifying) the List endpoints, 
https://github.com/chiefy/go-linode/compare/master...displague:more_resources_with_abstract_pagination?expand=1

With this, `ListStackscripts` would take a ListOptions which can contain a `Filter string`.
I haven't mocked out how the Filter string would be passed along, perhaps your approach makes the most sense (because of the complexity of the X-Filter parameters).

An alias of `ListAccountStackscripts` or `ListPublicStackscripts` would make sense too.

dvestal:
I apologize for the delay with my response. I have been dealing with a lot of issues regarding the new GDPR laws going into place. I will be happy to add the tests as soon as I can get back to work on this project. I will review the changes and suggestions @displague has within his generalization as well.

chiefy:
@dvestal no worries!

PR #11, opened 2018-05-02
Add pagination and read-support for other endpoints

### LINKED ISSUE #4: Implement godoc format docs
Need better inline documentation.

### LINKED ISSUE #5: Implement / expose ability to page through responses

### LINKED ISSUE #12: Rename repo
Per conversations with @displague we are going to rename the project to `linodego`

This change will require updating package refs.

displague:
This branch should pass tests and the new endpoints that require multiple IDs (nodebalancer + config, for example) should be verified before making it into master.

This PR of the pagination branch is best summarized by the PRs that went into pagination:
- https://github.com/chiefy/go-linode/pull/8 More resources with abstract pagination
- https://github.com/chiefy/go-linode/pull/10 add many new list and list with id endpoints to client

This PR sets up pagination through a `ListOptions` argument supplied to any `ListThing` call.

Closes #9, Closes #5, Closes #12, Closes #4

### Examples
Some `Example` tests (no API key required) were added to the `examples/` directory, run `go test` there. (these should be added to the regular tests later, for automatic godoc inclusion)
Similar examples are included in the notes below, see:
-  https://github.com/chiefy/go-linode/blob/pagination/example/example_test.go

### Without an API Key
Additionally, the normal tests can be run with `go test -test.short` to limit the tests to the ones that don't require an API key.  

### Recorded API interactions for tests
(some manual tweaking of client_test.go may be required until the latest fixtures are recorded:
```go
var testingMode = recorder.ModeReplaying
var debugAPI = false
var validTestAPIKey = "NOTANAPIKEY"
```

### Auto-Pagination Requests
```go
kernels, err := golinode.ListKernels(nil)
// len(kernels) == 218
```

### Single Page
```go
opts := NewListOptions(2,"")
// or opts := ListOptions{PageOptions: &PageOptions: {Page: 2 }}
kernels, err := golinode.ListKernels(opts)
// len(kernels) == 100
// opts.Results == 218
```
## Filtering
```go
opts := ListOptions{Filter: "{\"mine\":true}"}
// or opts := NewListOptions(0, "{\"mine\":true}")
stackscripts, err := golinode.ListStackscripts(opts)
```

### Getting Single Entities
```go
linode, err := golinode.GetLinode(555) // any Linode ID that does not exist or is not yours
// linode == nil: true
// err.Error() == "[404] Not Found"
// err.Code == "404"
// err.Message == "Not Found"
```

### Lists
For lists, the list is still returned as `[]`, but `err` works the same way as on the `Get` request.
```go
linodes, err := golinode.ListLinodes(NewListOptions(0, "{\"foo\":bar}"))
// linodes == []
// err.Error() == "[400] [X-Filter] Cannot filter on foo"
```

Otherwise sane requests beyond the last page do not trigger an error, just an empty result:
```go
linodes, err := golinode.ListLinodes(NewListOptions(9999, ""))
// linodes == []
// err = nil
```
### Writes
When performing a `POST` or `PUT` request, multiple field related errors will be returned as a single error, currently like:
```go
// err.Error() == "[400] [field1] the problem reported by the API
// [field2] the problem reported by the API
// [field3] the problem reported by the API"
```

### TODO
- [x] Make `Page=0` synonymous with `ListOptions=nil` (so filters can be supplied without breaking the auto-pagination)


chiefy:
Closes #12

PR #13, opened 2018-05-26
Privatize funcs

displague:
Make some internals private because these particular internals, related to URL fetching and pagination, could easily get refactored away.


chiefy:
:+1:

PR #14, opened 2018-05-29
Add Event polling

displague:
This PR:
- adds faux-enum strings for `EventAction`, event `EntityType`, and `EventStatus`
- adds some developers.linode.com copied field docs  
- adds Create, Update, Delete Domain (untested)
- adds Create, Update, Delete stubbing to template.go.example
- adds WaitForEventFinished which returns the Event when finished
- adds MarkEventRead and MarkEventsSeen 
- adds CreateStackscript, UpdateStackscript, DeleteStackscript
- example/main checks LINODE_SPEND and LINODE_DEBUG env vars to change behavior

I would like to find a way to offer `WaitForEventFinished` as a channel too..

I should probably split the Domain changes into another PR.

### DIFF
--- a/account_events.go
+++ b/account_events.go
@@ -9,20 +9,120 @@ import (
 // Event represents an action taken on the Account.
 type Event struct {
 	CreatedStr string `json:"created"`
-	UpdatedStr string `json:"updated"`
 
-	ID              int
-	Status          string
-	Action          string
+	// The unique ID of this Event.
+	ID int
+
+	// Current status of the Event, Enum: "failed" "finished" "notification" "scheduled" "started"
+	Status EventStatus
+
+	// The action that caused this Event. New actions may be added in the future.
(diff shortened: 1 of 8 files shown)


account_events.go
```diff
@@ -9,28 +10,129 @@ import (
...
+	Entity *EventEntity
+
+	// When this Event was created.
+	Created *time.Time `json:"-"`
 }
 
+// EventAction constants start with Action and include all known Linode API Event Actions.
+type EventAction string
+
+const (
```
tamalsaha: https://github.com/abice/go-enum is handy tool for generating enum for Go. We use this.

chiefy: thanks @tamalsaha will definitely check it out

displague: It's certainly interesting.  I think there is lots of room here to move towards generators.

PR #19, opened 2018-07-11
[wip] 429 handling

displague:
This is a work in progress branch to sort out 429 handling.

Some responses do not return the full 429 response, so special attention must be made to those more severe 429 responses.

Values like this may be returned on normal requests,
```
Retry-After: 106
X-Ratelimit-Limit: 400
X-Ratelimit-Remaining: 388
X-Ratelimit-Reset: 1531247494
```

And values like this, may be return on a 429 response:
```
TODO
```

And values like this, may be return on a 429 response that does not include all headers:
```
TODO
```


displague:
This PR should also handle `408 Please try again` responses in the same way -- delay with back-off, and retry.

PR #21, opened 2018-07-13
add SetBaseURL for use in non-production environments

displague:


displague:
I seem to have re-implemented this as `SetUrl` in https://github.com/chiefy/linodego/pull/27/files#diff-d0b7f86cea13f90d3cc0cc2338da8d5fR107

Any preference?

PR #36, opened 2018-08-07
Fixing Resource Pagination

cliedeman:
listHelperWithID is currently delegating to listHelper and hitting this [line](https://github.com/chiefy/linodego/blob/master/pagination.go#L264). Changing the fixture to return 2 pages exposes the error but I am not sure how to enhance the tests correctly

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -3,46 +3,57 @@
 
 [[projects]]
   branch = "master"
+  digest = "1:33ca3f9c42039adf6e511e0d761422020c90caf63db660f14e98a64d18b14524"
   name = "github.com/dnaeon/go-vcr"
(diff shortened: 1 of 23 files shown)


cliedeman:
Also, a log fatal is probably not good enough there. We should return an error.

displague:
Nice catch! Can you pick up https://github.com/chiefy/linodego/pull/36/files#diff-8772883845372b18b10f83b47f46c826R445 while in here?

I think the "correct" way to handle the fixture in this case, would be to make a new test that represents a multiple pages response (copy the old existing fixture to this new fixture) and hand edit the response so that multiple pages or responses are returned, as the subsequent result of a mocked ?page=2 request.

Issue #29, opened 2018-07-25
template.go needs to sync up with other changes


displague:
- [ ] update template.go to include Create/Update options
  - [ ] update template.go to include instructions
     For file:
     > replace "Template" with "Domain" or "DomainRecord", (likewise for lowercase) ...

     For types:
     > use pointers where ever `null`'able fields whose `0` or `""` values would have a different meaning to the API
  - [ ] update template.go to include [UpdateOptions and CreateOptions funcs](https://github.com/chiefy/linodego/blob/master/nodebalancer_configs.go#L99-L135)
 - [ ] mention that client.go and resources.go need to new items for each added Thing
- [ ] template.go should have commented out examples of the code sub resources would use

displague:
Closed by aaff5e8

PR #38, opened 2018-08-09
consistent pointer use in Create/Update calls

### LINKED ISSUE #37: Standardize Input Argument Pointers vs Values
Follow up to #35.

We need to decide on a standard for input arguments. Right now I am leaning towards everything being values?

displague:
- Adds CreateImage, UpdateImage, DeleteImage
- Changes UpdateOptions and CreateOptions and similar Options parameters to values instead of pointers, these were never optional and the function never updated any values in the Options structures
- Changes InstanceUpdateOptions to use pointers for optional fields Backups and Alerts, adds WatchdogEnabled
- Changes InstanceClone's Disks and Configs to ints instead of strings
- Adds RescueInstance and RescueInstanceOptions
- Updates template.go which can be used as a manual template when adding support for missing endpoints

This closes #37 and addresses most of the concerns in #29

### DIFF
--- a/API_SUPPORT.md
+++ b/API_SUPPORT.md
@@ -20,7 +20,7 @@
 - `/linode/instances/$id/rebuild`
   - [X] `POST`
 - `/linode/instances/$id/rescue`
-  - [ ] `POST`
+  - [X] `POST`
 - `/linode/instances/$id/resize`
   - [x] `POST`
 - `/linode/instances/$id/shutdown`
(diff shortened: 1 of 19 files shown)


displague:
I'm willing to take the coverage hit for now, because with https://github.com/chiefy/linodego/issues/32 we will create Instances and walk all of the config and disk functions in a controlled environment, without magic and the requirement that personal accounts have fixture conditions.

PR #46, opened 2018-08-30
Change ListOptions `Filter` field to map[string]interface{}

maddie:
This should make the filter much easier to use than plain `string` type. In the old API library I've written in a private project, I've used this same approach and it works quite well for me.

I think we can use Filters this way until Linode's v4 API becomes stable and/or we found a better way to handle Filters.

### DIFF
--- a/domain_records_test.go
+++ b/domain_records_test.go
@@ -2,7 +2,6 @@ package linodego_test
 
 import (
 	"context"
-	"encoding/json"
 	"testing"
 
 	"github.com/linode/linodego"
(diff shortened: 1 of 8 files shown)


displague:
When introducing ListOptions Filters, I was hesitant about `map[string]interface{}` because I thought the format could have changed, or be amended to allow for a simpler alternative: 
```
X-Filter: (simpler eq true)
```

But, I've confirmed that there are no plans to change `X-Filter` behavior, and in fact, the API docs back this up:

> Filters are passed in the X-Filter header and are formatted as JSON objects

---

If we accept `interface{}` here, we could continue to accept `string`, and `map[string]interface{}`, and perhaps some other functional filter builder.

Another option could be to accept anything that implements some JSONable interface: https://play.golang.org/p/JiGhwIRDxOI (there must be an existing type for this? right?)

That may be overkill though.  `map[string]interface{}` is probably what we actually want.

maddie:
Using `interface{}` might actually make `ListOptions` unwieldy and possibly unclear for users, since they can't be sure of what kind of data should they pass in and could fail in some weird way. Also, using `interface{}` would mean extra validation for `X-Filter`s passed in might be required in the library.

jriddle-linode:
closing outdated

PR #48, opened 2018-09-03
Adding Nodebalancer rebuild

### LINKED ISSUE #47: Implement API 4.0.4 changes (mostly SSH Key support)
Per https://developers.linode.com/changelog

- [x] Add profile_sshkeys.go, profile_sshkeys_test.go (use template.go)
  - [x] Revise RebuildInstance and InstanceRebuildOptions
  - [x] Revise CreateInstance and InstanceCreateOptions
  - [x] Revise CreateDisk and InstanceDiskCreateOptions
- [x] Implement RebuildNodebalancerConfig
- [x] Revise CreateNodebalancer for fast-track config settings (from v4.0.2)

cliedeman:
Closes #47


nodebalancer_configs.go
```diff
@@ -103,6 +103,33 @@ type NodeBalancerConfigCreateOptions struct {
 	SSLKey        string           `json:"ssl_key,omitempty"`
 }
 
+// NodeBalancerConfigRebuildOptionsNode for a nodebalancer instance mapping
+type NodeBalancerConfigRebuildOptionsNode struct {
```
displague: This is very similar to `NodeBalancerNodeCreateOptions`

https://github.com/linode/linodego/blob/master/nodebalancer_config_nodes.go#L35

cliedeman: It was copied from there :). I could not actually verify if the inputs are the same because the new api's do not seem to be in the [Open API spec](https://developers.linode.com/openapi.yaml)

nodebalancer_configs.go
```diff
@@ -103,6 +103,33 @@ type NodeBalancerConfigCreateOptions struct {
...
+	CheckInterval int                                     `json:"check_interval,omitempty"`
+	CheckAttempts int                                     `json:"check_attempts,omitempty"`
+	CheckPath     string                                  `json:"check_path,omitempty"`
+	CheckBody     string                                  `json:"check_body,omitempty"`
+	CheckPassive  *bool                                   `json:"check_passive,omitempty"`
+	CheckTimeout  int                                     `json:"check_timeout,omitempty"`
+	CipherSuite   ConfigCipher                            `json:"cipher_suite,omitempty"`
+	SSLCert       string                                  `json:"ssl_cert,omitempty"`
+	SSLKey        string                                  `json:"ssl_key,omitempty"`
+	Nodes         []*NodeBalancerConfigRebuildOptionsNode `json:"nodes,omitempty"`
```
displague: I'm not sure what the behavior is for a rebuild with an empty node array (`"nodes":[]`).   

If it removes all of the nodes (any differently than not specifying `"nodes"` in the json) we may want to avoid `omitempty` to allow one to clear the nodes.

---

Likewise, `CheckInterval`, `CheckTimeout`, `CheckAttempts`, `Port` may have meaningful `0` values to consider.
Likewise, `CheckPath`, `CheckBody`, may have special meaning as `""`.

This probably lines up with NodeBalancerConfigCreateOptions, and if so I would hope we could just alias the type.

PR #52, opened 2018-09-15
Testing go mod vendor

cliedeman:


ciaran-stackworx:
I believe the old linter was the main reason I needed the go mod vendor. Hopefully it can be removed now

PR #45, opened 2018-08-30
Go 1.11 modules support

maddie:
Fixes #44

PR #53, opened 2018-09-16
Switch go-resty import path to gopkg.in version for future go module …

cliedeman:
Until #45 is resolved this should let other go mod projects consume linodego without issue.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -13,14 +13,6 @@
   revision = "aafff18a5cc28fa0b2f26baf6a14472cda9b54c6"
 
 [[projects]]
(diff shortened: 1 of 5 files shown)


kzap:
@cliedeman @displague until a new version is released of this package users will need to add the sha1 to their `go.mod` file to make sure they get this version of the package:

```go
require (
	github.com/linode/linodego 15d4ab6c221bdc0a02f1e56c45199705cd4c9ed5
)
```

cliedeman:
You can use use master instead of a version tag and go mod will lock to the latest commit available on master. Have used it on a few projects already. Just a workaround until the next version is released

PR #55, opened 2018-09-27
Fix broken example code in README

kzap:
The current client example is broken because of a missing `net/http` package and `linodego.NewClient` returns only 1 variable, not 2

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,13 +37,15 @@ A brief summary of the features offered in this API client are shown here.
 package main
 
 import (
-  "context"
-  "fmt"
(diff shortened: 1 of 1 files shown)


displague:
Thanks.  Documentation, code that doesn't get executed or tested.

There are examples that are actually tested here, but nothing that shows the NewLinodeClient function in action, since the examples/tests use a test client:
https://godoc.org/github.com/linode/linodego#pkg-examples

kzap:
thanks, was looking for those examples 👍

PR #61, opened 2018-11-27
fix 404 link for v4 API docs

carlwgeorge:


displague:
Thanks, @carlwgeorge!

PR #72, opened 2019-03-30
Add GET linode instance stats

MichaelSaah:
First go adding linode instance stats to the client. Needs tests. Names likely suck. Will probably try to get other stats in as well on this PR.

Would be nicer if datapoints were unmarshalled into a struct of {int,float}, since the first value looks like an epoch time stamp in milliseconds. Currently just unmarshalling into []float64.

Implements https://developers.linode.com/api/docs/v4#operation/getLinodeStats


displague:
Thanks @MichaelSaah! 

This return format is also used by https://developers.linode.com/api/docs/v4#operation/getLinodeStatsByYearMonth which will be easy to add support for following this PR.

From the documentation, it doesn't appear that IO, NetV4, and NetV6 can ever be `nil`, nor can `Data`.  For struct members that will never return `nil`, the convention has been to return the struct and not a reference to it.  This bit of early sanity enforcement prevents users from needing to check if the member exists before accessing it.

In terms of names, I would avoid bare names since they share the same namespace as everything else in linodego currently.

* `InstanceStatsData`
* `InstanceStatsIO` (or `StatsIO`)
* `InstanceStatsNetV4` (or `StatsNetV4`)
* `InstanceStatsNetV6` (or `StatsNetV6`)

I went looking for existing convention and the closest I could find was `InstanceSpecs` and `NodeBalancerTransfer`, I don't think there is an existing parallel.

MichaelSaah:
Started writing a test but ended up confused. Can you explain how tests are supposed to work with fixtures?

displague:
Explained this offline, essentially: https://github.com/linode/linodego#tests

`make ARGS="-run TestNewFeature" fixtures` will create a file with captured request and response pairs from the actual API calls.  This file, which will be created in the `fixtures/` directory, should be scrutinized for sensitive values and added to the PR (there is a scrubber in place, but double check).

Subsequent invocations of `make test` use the fixtures to quickly test the code without reaching out to the real Linode API.

The tests should therefore not make any assumptions about the resources that are available on a given account.   The tests should create new resources, put them in the expected state for testing, and delete the resources when done.

MichaelSaah:
Wrote a test, but ran into a problem. It seems that stats don't become available for a linode immediately after first boot.

Spun up a nanode in newark and it took about 5m 30s to start returning stats. If you're OK with the test sleeping for 10 minutes or so, then I'll go with that.

If you can think of a better solution, please suggest.

MichaelSaah:
Well it worked, used a 9 minute sleep period, since it looks like the testing lib kills tests that run >10 minutes.

MichaelSaah:
looks like the test passed, failed on linter stuff. I'll fix that.

MichaelSaah:
🎉

PR #80, opened 2019-06-05
Add support for alternate root CA certs

jfrederickson:


asauber:
@jfrederickson can you determine if we can write a test for this?

Otherwise, I'd like to merge it

PR #81, opened 2019-06-06
add new PersistAcrossBoots flag for volume attachments

displague:
Adds support for `persist_across_boots`:

https://developers.linode.com/api/docs/v4#operation/attachVolume

This is needed for CSI driver support beyond 8 devices, for example.

https://www.linode.com/community/questions/17538/linode-block-storage-csi-driver#answer-69053

### DIFF
--- a/volumes.go
+++ b/volumes.go
@@ -50,7 +50,8 @@ type VolumeCreateOptions struct {
 	// The Volume's size, in GiB. Minimum size is 10GiB, maximum size is 10240GiB. A "0" value will result in the default size.
 	Size int `json:"size,omitempty"`
 	// An array of tags applied to this object. Tags are for organizational purposes only.
-	Tags []string `json:"tags"`
+	Tags               []string `json:"tags"`
+	PersistAcrossBoots *bool    `json:"persist_across_boots,omitempty"`
 }
 
 // VolumeUpdateOptions fields are those accepted by UpdateVolume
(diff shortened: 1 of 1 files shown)


jnschaeffer:
Should this also include tests in `volumes_test.go`?

volumes.go
```diff
@@ -61,8 +62,9 @@ type VolumeUpdateOptions struct {
 
 // VolumeAttachOptions fields are those accepted by AttachVolume
 type VolumeAttachOptions struct {
-	LinodeID int `json:"linode_id"`
-	ConfigID int `json:"config_id,omitempty"`
+	LinodeID           int   `json:"linode_id"`
+	ConfigID           int   `json:"config_id,omitempty"`
```
jnschaeffer: My only gripe with this is that a `ConfigID` of 0 is semantically valid, if (probably) not valid in practice. Is there a reason we don't use `*int` here?

asauber: These ConfigIDs are large unique integers (across all Linodes), so I don't think this is necessary.

displague: It is not a required field so we `omitempty`, I don't recall if the LinodeAPI errors on `ConfigID:0` or if it otherwise has the same effect as not specifying the argument at all.  IIRC, without `PersistAcrossBoots` the LinodeAPI behavior is to  match this volume to the "current" Linode Config when ConfigID is not sent.

I would say I generally kept the user from having to work with `*int` unless it couldn't be avoided because absence (default behaviors), `nil`, and/or `0` had a different meaning to the Linode API.

Also, that would have been a breaking change which this PR does not otherwise include.

displague: If the api doesn't accept 0 then there is no reason to send it.  I have followed the documentation as closely as possible for these.

https://developers.linode.com/api/docs/v4#operation/attachVolume
> The ID of the Linode Config to include this Volume in. Must belong to the Linode referenced by linode_id. If not given, the last booted Config will be chosen.

* Config "0" does not belong to the linode_id. it is not valid.
* "not given" is the behavior we must be able to achieve

PR #88, opened 2019-06-25
Release v0.10.0

asauber:
### Breaking Changes

* Change to `AllowAutoDiskResize` to `*bool` is a breaking change, so bump minor version.


asauber:
Thank you @displague going to force update that before people start using the tag

PR #89, opened 2019-07-12
[WIP] new: Added Listener and InstanceFuture types

jnschaeffer:
This is a sketch of how we can start moving away from functions that potentially poll the API many times waiting for statuses of Linode instances, events, or any other things that we need to poll for. It doesn't currently integrate with anything else in `linodego`, but @ellisbenjamin and I were talking about this so why not show some work?

### DIFF
--- a/listener.go
+++ b/listener.go
@@ -138,8 +138,10 @@ func (s *Listener) getResults(ctx context.Context) error {
 	}
 
 	// We could wrap each send in a goroutine, but that might just make it harder to track down
-	// resource leaks later.
-	for _, instance := range instances {
+	// resource leaks later. Since we need to take references to individual items in the slice,
+	// we just iterate over the index here.
+	for i := range instances {
+		instance := &instances[i]
 
 		// Bail if we somehow find an instance with no associated future.
 		future, ok := s.instanceFutures[instance.ID]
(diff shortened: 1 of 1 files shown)


displague:
The aggregated filters that this polling system introduces should prevent 429s and improve overall responsiveness.

This raises a few questions:
* would existing `WaitFor*` functions be retro-fitted to use a shared Poll behind the scenes
* typical Event filtering (by newness) is different than instance filtering (by id), an intelligent base filter (use the min age of all event filters) may be necessary to optimize shared event polling.
* does this warrant a `futures` sub-package?
* could this be refactored in such a way that each Linode Thing implements what is necessary for the poller, rather than requiring a new customized copy of listerner.go for  each Thing.
* how could back-off be optimized using a shared polling? Should the filter perhaps return (instead of a bool) a signal to increase or decrease the polling interval (a low `percent_complete` value returned from an event, or a `creating` status from a Linode Instance may indicate that the polling can back-off a little while, while a nearly complete percent or late stage status could yield a signal to decrease the polling interval.)

displague:
An example of usage in the PR description would be helpful, contrasting how multiple WaitFor's would be necessary now compared to the benefits of this design.

phillc:
This appears to have aged beyond context. If this is still relevant today, please do restart this conversation.

PR #90, opened 2019-07-22
Preliminary support for object storage

ellisbenjamin:
This supports listing clusters, and CRUD ops for keys.

Need to set LINODE_URL="https://api.linode.com/v4beta" for this to work.


displague:
Based on the API end-points,
```
	clustersEndpoint            = "object-storage/clusters"
	objkeysEndpoint             = "object-storage/keys"
```

I would offer that the current naming convention would have these referred to as:
```
*ObjectStorageKey
*ObjectStorageCluster
```

or 

```
*StorageKey
*StorageCluster
```

I don't think the following would fit the current convention. 

```
*OSKey
*OSCluster

* ObjKey
* Cluster
```

I also wonder about how obvious `Cluster` and `ObjKey` would be to users.

displague:
https://github.com/linode/linodego/blob/master/API_SUPPORT.md should also be updated with the new endpoints (and doing so should probably be added to the template.go file so it isn't missed in the future).

ellisbenjamin:
@displague re: naming I think the linodego client has been a little inconsistent here unless there's a convention that I'm missing:

eg, we have `invoicesEndpoint            = "account/invoices"` rather than `accountInvoiceEndpoint`

but then we also have 

`accountSettingsEndpoint     = "account/settings"` where we might instead have `settingsEndpoint`

and..

`eventsEndpoint              = "account/events"`

and..

`oauthClientsEndpoint        = "account/oauth-clients"`

In no cases do we shorten or abbreviate the resource name in the manner I'm doing here from `objectstorage` to `obj`. In any case, I'm in favor of making names as unambiguous as possible and your point is well taken.

I'm going to make the appropriate changes and resubmit.

From what I understand, we may be dropping `clusters` in favor of `regions` at some point in the not-to-distant future, which would be nice.

ellisbenjamin:
@displague  -- I'm in agreement with you regarding the naming convention. I've updated this PR accordingly, and have updated template.go and API_SUPPORT.md also.

phillc:
Superseded by #95

Issue #84, opened 2019-06-18
panics occurring when API is down


jackmiller334:
Hi. 

We had some panics using this package coinciding with the recent Linode API service issues. 
https://status.linode.com/incidents/ych41x06fm8k

```
Jun 18 01:10:40 023c8ee9f52b runtime.log: panic: runtime error: invalid memory address or nil pointer dereference
Jun 18 01:10:40 023c8ee9f52b runtime.log: [signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0x865f9a]
Jun 18 01:10:40 023c8ee9f52b runtime.log: goroutine 43 [running]:
Jun 18 01:10:40 023c8ee9f52b runtime.log: github.com/linode/linodego.(*Client).listHelper(0xc000338000, 0xbcb4e0, 0xc000024090, 0xa26300, 0xc00049da40, 0xc00065d660, 0x9e40e0, 0xc000211301)
Jun 18 01:10:40 023c8ee9f52b runtime.log:     /app/src/github.com/linode/linodego/pagination.go:91 +0x5fa
Jun 18 01:10:40 023c8ee9f52b runtime.log: github.com/linode/linodego.(*Client).ListRegions(0xc000338000, 0xbcb4e0, 0xc000024090, 0xc00065d660, 0xc0004dd5a8, 0xfff2a8, 0xc0002c0900, 0x0, 0x0)
Jun 18 01:10:40 023c8ee9f52b runtime.log:     /app/src/github.com/linode/linodego/regions.go:37 +0x78
Jun 18 01:10:40 023c8ee9f52b runtime.log: main.(*LinodeHost).getRegions(0xc0002ac390, 0xc000397500, 0x9, 0x11, 0x0, 0x0)
```

```
Jun 18 00:38:57 023c8ee9f52b runtime.log: panic: runtime error: invalid memory address or nil pointer dereference
Jun 18 00:38:57 023c8ee9f52b runtime.log: [signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0x8699f2]
Jun 18 00:38:57 023c8ee9f52b runtime.log: goroutine 101 [running]:
Jun 18 00:38:57 023c8ee9f52b runtime.log: github.com/linode/linodego.(*Client).listHelper(0xc0003b4000, 0xbcb4c0, 0xc000024090, 0xa25b00, 0xc000530f60, 0xc0005611d8, 0x48ae19, 0xc0005610d8)
[...]

PR #91, opened 2019-09-25
Check Content-Type before unmarshalling json

adammohammed:
This addresses the issue #84. When the API servers were returned a 502, linodego panicked. This happened due to the client trying to parse the response as json when a `text/html` response was received.

The solution is to check the response content-type to see if it matches the content-type accepted by the request. All of the code base assumes that the response is well-formed json, so it may be okay to just check to see if the content-type matches "application/json".

If you remove the code added in `errors.go` the included test will fail due to panicking when trying to unmarshal the html response.

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -49,6 +49,14 @@ func coupleAPIErrors(r *resty.Response, err error) (*resty.Response, error) {
 	}
 
 	if r.Error() != nil {
+		// Check that response is of the correct content-type before unmarshalling
+		expectedContentType := r.Request.Header.Get("Accept")
+		responseContentType := r.Header().Get("Content-Type")
+		if responseContentType != expectedContentType {
+			msg := fmt.Sprintf("Expected: %v; Received: %v", expectedContentType, responseContentType)
+			return nil, NewError(msg)
+		}
+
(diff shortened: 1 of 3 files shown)


errors.go
```diff
@@ -49,6 +49,14 @@ func coupleAPIErrors(r *resty.Response, err error) (*resty.Response, error) {
 	}
 
 	if r.Error() != nil {
+		// Check that response is of the correct content-type before unmarshalling
+		expectedContentType := r.Request.Header.Get("Accept")
+		responseContentType := r.Header().Get("Content-Type")
+		if responseContentType != expectedContentType {
+			msg := fmt.Sprintf("Expected: %v; Received: %v", expectedContentType, responseContentType)
```
jnschaeffer: It'd probably be better if this error message was something like `unexpected content type: expected %v, got %v` and if the test above checked for that.

adammohammed:
@jnschaeffer Test was added to check the error does get properly raised, and that the message involves content-type. 
cc. @phillc

phillc:
I will bump linodego version after I got the OBJ endpoints in.

PR #95, opened 2019-09-27
Object Storage

phillc:


object_storage_clusters_test.go
```diff
@@ -0,0 +1,19 @@
...
+	client, teardown := createTestClient(t, "fixtures/TestListObjectStorageClusters")
+	defer teardown()
+
+	objectStorageClusters, err := client.ListObjectStorageClusters(context.Background(), nil)
+	if err != nil {
+		t.Errorf("Error listing objectStorageClusters, expected struct - error %v", err)
+	}
+	if len(objectStorageClusters) == 0 {
+		t.Errorf("Expected a list of objectStorageClusters - %v", objectStorageClusters)
+	}
```
jnschaeffer: Do we want to assert that the content of the cluster list is what we expect?

PR #105, opened 2019-11-25
adds CRUD support for LKE and LKE NodePool endpoints

displague:
Adds support for CRUD operations on LKE endpoints per, https://developers.linode.com/api/v4/lke-clusters.

Adds more instructions to `template.go` concerning generating test fixtures and removing the `ignore` build directive.

This PR has opted to call the new resources `LKECluster` and `LKEClusterPool`.  Let me know if `LKEClusterNodePool` or other names would offer a better UX.

Some caveats were encountered with the beta API:
* LKE Clusters can not be filtered by label, as documented. This limits the accuracy of the ListLKECluster test which would have looked for the precise cluster by label.
* LKE Cluster deletes are not quick enough for a create with the same label to be created.  A setup modifier pattern was added in these tests to tweak the label per test. This original pattern of a randomized label was taken from the nodebalancer tests, which haven't run into the problem of label reuse.
* https://developers.linode.com/api/v4/tags-label docs do not specify what the `types` are.  I had to guess that `lke_clusters` would be returned by the Tag endpoints.  This guess is likely wrong.
* The API docs for creating node pools does not detail that `type` is a `string`.  I am assuming that it is (which passes tests).
* ~The API returns `"1234"` and `""` for the `id` in a pool's `linodes` list.  This is documented as an `int` or `null`. _I've notified Linode of this bug and worked around this with a `fixID` method._~ This API problem has been fixed, so the workaround has been removed from this PR.

---
Changes:

* adds `LKEClusters` to tag lists
* add `CreateLKECluster`, `GetLKECluster`, `ListLKEClusters`, `UpdateLKECluster`, `DeleteLKECluster`, `WaitForLKEClusterReady`
* add `CreateLKEClusterPool`, `GetLKEClusterPool`, `ListLKEClusterPools`, `UpdateLKEClusterPool`, `DeleteLKEClusterPool`, `WaitForLKEClusterPoolReady`
* add `ListLKEVersions`, `GetLKEVersion`, `GetLKEClusterAPIEndpoint`, `GetLKEClusterKubeconfig`
* add `LKEClusterReady`, `LKEClusterNotReady` enums of type `LKEClusterStatus`
* add `LKELinodeReady`, `LKELinodeNotReady` enums of type `LKELinodeStatus`

----

- `/lke/clusters`
  - [X] `POST`
  - [X] `GET`
  - [X] `PUT`
  - [X] `DELETE`
- `/lke/clusters/$id/pools`
  - [X] `POST`
  - [X] `GET`
  - [X] `PUT`
  - [X] `DELETE`
- `/lke/clusters/$id/api-endpoint`
  - [x] `GET`
- `/lke/clusters/$id/kubeconfig`
  - [x] `GET`
- `/lke/clusters/$id/versions`
  - [x] `GET`
- `/lke/clusters/$id/versions/$id`
  - [x] `GET`


lke_clusters.go
```diff
@@ -0,0 +1,266 @@
...
+	}
+	e = fmt.Sprintf("%s/%s", e, version)
+	r, err := coupleAPIErrors(c.R(ctx).SetResult(&LKEVersion{}).Get(e))
+	if err != nil {
+		return nil, err
+	}
+	return r.Result().(*LKEVersion), nil
+}
+
+// ListLKEClusters lists LKEClusters
```
abemassry: Should be:
`// ListLKEVersions lists LKEVersions `

displague:
Thanks, @abemassry - addressed in https://github.com/linode/linodego/compare/29c83c1662bb47da4390af7aaa3ee33035316bc5..9a5a6d7265eed7a766c018bc91e9a66512146867

DazWilkin:
FWIW, I tested this with some code and it works (well).

The only change I had to make was in `Client.go` and `APIVersion = "v4beta"`

displague:
@DazWilkin that makes sense.

There is an environment variable that can be used to avoid those code changes.

`LINODE_API_VERSION=v4beta`

@phillc some documentation about this variable in README.md could be useful, especially as a few endpoints have relied on this feature.

phillc:
Thank you for your outstanding contribution.

PR #70, opened 2019-02-04
429 and 408 handling

displague:
This is a work in progress branch to sort out 429 handling. (Replaces #19 whose source branch is gone)

* [x] Needs tests
      - complicated because the current test runner does not handle repeated requests
      - a `[ GET /something [429], GET /something [200] ]` list of fixtures would be keyed on the METHOD+URL and so the `[429]` response will loop

Some responses do not return the full 429 response, so special attention must be made to those more severe 429 responses.

Headers like this may be returned on normal requests,
```
Retry-After: 106
X-Ratelimit-Limit: 400
X-Ratelimit-Remaining: 388
X-Ratelimit-Reset: 1531247494
```

And headers and body values like this, may be return on a 429 response:
```
TODO provide detail snippet here
```

And headers and body values like this, may be return on a 429 response that does not include all headers:
```
TODO provide detail snippet here
```

And 408 headers and body messages may be returned:
```
TODO provide detail snippet here
```

PR #107, opened 2019-12-02
Retry on Linode Busy errors.

phillc:


displague:
@phillc you may be interested in comparing notes with https://github.com/linode/linodego/pull/70/files.  The fixture in that PR was hand-crafted because generating this fixture dynamically was not guaranteed to be reproducible.

client.go
```diff
@@ -154,10 +154,30 @@ func (c *Client) SetToken(token string) *Client {
 	return c
 }
 
+func (c *Client) SetLinodeBusyRetry() *Client {
```
displague: drive-by comments..


`Linode` may be redundant naming here.  OTOH, `SetAPI*` is patterned above.

A type that defines the retry values could be useful, and this function could take a reference to a variable of that type.

This function should have  a go doc and it should call out the behavior  `SetPollDelay` has on this.

phillc: Point 1, I would have named the method "Server Busy" if the error message said that. Its the only error I am targeting right now.

I understand your point on the second. I will keep this in mind for refactor in the future.

Addressed point 3

PR #116, opened 2020-01-23
Split test suite into unit vs integration

phillc:


account_events.go
```diff
@@ -4,10 +4,10 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"log"
-	"strconv"
-	"strings"
 	"time"
+
+	"github.com/linode/linodego/internal/parseabletime"
+	"github.com/linode/linodego/internal/utils"
```
displague: https://blog.golang.org/package-names#TOC_5.

phillc: good call

displague:
How does renaming the test package affect the examples?
https://godoc.org/github.com/linode/linodego#pkg-examples

phillc:
> How does renaming the test package affect the examples?
> https://godoc.org/github.com/linode/linodego#pkg-examples

Are you asking why the tests failed? Or are you saying there is an oversight on the package rename?

PR #117, opened 2020-01-25
Provide LKE event actions

sgmac:
I'm working on the `terraform-provider-linode` to add the LKE feature. However, I found out there are not any Event Action for LKE.  Therefore, operations create and delete wait for indefinitely or exit right away. 

```go
client.WaitForEventFinished(context.Background(), id, linodego.EntityLinode, linodego.ActionLKEDelete, minDelete, int(d.Timeout(schema.TimeoutDelete).Seconds()))
```
I saw that in https://developers.linode.com/api/v4/account-events there is a list of events,
but I didn't see any definition for this. I also tried on `v4beta` without any luck. Maybe I'm missing  something.


displague:
@sgmac `template.go` should get a bullet reminding people to add the new events. I missed this in #105.

displague:
@sgmac I don't see these events documented here: https://developers.linode.com/api/v4/account-events

Did you see them somewhere else?

sgmac:
@displague  That's the main issue, there is nothing defined yet. I reached out to the team on Slack, it seems the addition of events for LKE is still on-going.

jnschaeffer:
Hi @sgmac! You're right that we don't provide information about the status of the cluster via Linode API events currently.

When creating an LKE cluster, there are a few points where you could consider the cluster to be ready:
- The API server becoming available (docs: https://developers.linode.com/api/v4/lke-clusters-cluster-id-api-endpoint)
- The kubeconfig for administering the cluster becoming available (docs: https://developers.linode.com/api/v4/lke-clusters-cluster-id-kubeconfig)
- All machines in all node pools running (docs: https://developers.linode.com/api/v4/lke-clusters-cluster-id-pools)
- All nodes in all node pools becoming available
- All control plane pods and services becoming available on all nodes

Because the cluster's API server and kubeconfig become available well before any nodes or pods do (i.e. within a few seconds of launching the request), you could get the kubeconfig for the cluster being provisioned and check for the status of nodes via the Kubernetes API, using `kubectl` or another utility like the Kubernetes Go client. This means that you can start interacting with your cluster almost immediately and submitting resources to it before nodes become available.

There are `WaitFor` functions, such as `WaitForInstanceStatus` and `WaitForClusterStatus`, that you could use to check for the status of node pool nodes. Another Kubernetes API-based solution might involve waiting for the nodes, but this comes with a caveat: `kubectl wait nodes` won't behave as expected if your cluster's nodes don't exist yet.

All of this is to say that the readiness of a cluster depends on many factors; the most basic and easiest to check in practice would probably be that the cluster's API server is available and that the kubeconfig you get from the Linode API works when connecting to the cluster.

I hope this is helpful information! Thanks for your contributions - looking forward to seeing what you come up with 👍

displague:
#143 includes the definition of the `ActionLKENodeCreate` event constant, the only event that the [API docs for events](https://developers.linode.com/api/v4/account-events#list-events) report.

sgmac:
Then I'll close this PR. Thanks !

PR #121, opened 2020-02-11
Add SetRetryAfterRateLimit to Client

jnschaeffer:
`SetRetryAfterRateLimit` sets the client to retry on 429 responses so linodego is able to more gracefully handle API rate limits.


phillc:
Rejected in favor of internal refactor.

phillc:
Resty backs off by default.

PR #122, opened 2020-02-12
make Makefile more resilient

Charliekenney23:


Makefile
```diff
@@ -1,4 +1,4 @@
-include .env
+-include .env
```
Charliekenney23: Should not exit for missing `.env` file. Not everyone will be running integration tests

Makefile
```diff
@@ -25,7 +25,12 @@ vet:
 	go vet ./...
 
 lint:
+ifndef $(shell command -v $(GOLANGCILINT) -v dot 2> /dev/null)
```
Charliekenney23: Warn user about missing golangci-lint installation, but don't exit.

sarahelizgray: Why not exit? The linting step seems useful, especially for outside contributors.

Charliekenney23: My hope was that this warning would provide an easier way to call to action actually installing golangci-lint on the users machine nstead of:
```
make: golangci-lint: No such file or directory
```
But I see where you're getting at. Would you prefer that we log the installation instructions and then exit?

PR #125, opened 2020-02-15
WaitForLKEClusterPoolStatus feature

sgmac:
While working in adding the LKE feature to the `terraform-provider-linode` I noticed that `WaitForLKEClusterStatus` returns immediately. However, this does not guarantee the LKE cluster is in ready status. The LKE Cluster is operational when all the instances in each cluster pool defined, are in ready status. 

I added a test just using this new function, it may be overkilling as it's just like `TestGetLKECluster_found`.

By adding this function, now the create LKE cluster on Terraform is more realistic:

```shell
linode_lke_cluster.beta: Still creating... [1m10s elapsed]
linode_lke_cluster.beta: Still creating... [1m20s elapsed]
linode_lke_cluster.beta: Still creating... [1m30s elapsed]
linode_lke_cluster.beta: Still creating... [1m40s elapsed]
linode_lke_cluster.beta: Creation complete after 1m48s [id=1559]

Apply complete! Resources: 1 added, 0 changed, 0 destroyed.
```

This PR includes  #124 , I need those changes to test this functionality.

### DIFF
--- a/waitfor.go
+++ b/waitfor.go
@@ -187,6 +187,46 @@ func (client Client) WaitForLKEClusterStatus(ctx context.Context, clusterID int,
 	}
 }
 
+// WaitForLKEClusterPoolStatus determines if a given LKE cluster is ready
+// only when all the instances of the cluster pool are in ready status.
+func (client Client) WaitForLKEClusterPoolStatus(ctx context.Context, clusterID int, timeoutSeconds int) error {
+	ctx, cancel := context.WithTimeout(ctx, time.Duration(timeoutSeconds)*time.Second)
+	defer cancel()
+
+	ticker := time.NewTicker(client.millisecondsPerPoll * time.Millisecond)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ticker.C:
+			pools, err := client.ListLKEClusterPools(ctx, clusterID, nil)
+			if err != nil {
+				return fmt.Errorf("Error retrieving the LKE Cluster Pool associated to %d %s", clusterID, err)
+			}
+
+			allNodesReady := func(p []LKEClusterPool) bool {
+				var tNodes int
+				for _, pool := range p {
+					tNodes = tNodes + pool.Count
+					for _, node := range pool.Linodes {
+						if node.Status == LKELinodeReady {
+							tNodes = tNodes - 1
+						}
+					}
+				}
+				return tNodes == 0
+			}(pools)
+
+			if allNodesReady {
+				return nil
+			}
+
+		case <-ctx.Done():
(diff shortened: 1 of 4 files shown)


waitfor.go
```diff
@@ -187,6 +187,46 @@ func (client Client) WaitForLKEClusterStatus(ctx context.Context, clusterID int,
...
+		case <-ticker.C:
+			pools, err := client.ListLKEClusterPools(ctx, clusterID, nil)
+			if err != nil {
+				return fmt.Errorf("Error retrieving the LKE Cluster Pool associated to %d %s", clusterID, err)
+			}
+
+			allNodesReady := func(p []LKEClusterPool) bool {
+				var tNodes int
+				for _, pool := range p {
+					tNodes += pool.Count
```
displague: I originally included a function like this in the LKE PR but replaced it with WaitForLKEClusterStatus because it matched the API provided but also because philosophically it shouldn't technically matter if nodes are available before pushing work to the cluster.

I see the value in this function, regardless.

I would avoid the counting up and down and simply return false if any nodes are not ready.

Alternatively, this function could take an availability goal.  1% would indicate (in most use-cases) that a single node is present.  100% would request that all nodes be available. Constants could be provided for various meaningful goals (>=1/1000, >= 1/2, >= 2/3, =100).

sgmac: Thanks! I appreciate your suggestions, I'll give it a try and implement either of them.

displague: > because philosophically it shouldn't technically matter if nodes are available before pushing work to the cluster

Adding to this justification for not necessarily needing these `WaitForLKECluster*` function, when the Kubernetes API is available, `kubectl get nodes` (https://...k8s-api.../api/v1/nodes) will be a better indicator.

sgmac: I think the closes functionality to this would be to check for when the `kubeconfig` is ready. I could use GetLKEClusterKubeconfig, at least this way does not return immediately. WDYT? I see this having in mind the `terraform-provider`.

displague: As it relates to the terraform provider, I'm not really sure, @sgmac.

I would defer to @asauber @jfrederickson @jnschaeffer and @phillc as the authorities on best-practice here.

https://developers.linode.com/api/v4/lke-clusters-cluster-id-kubeconfig does not suggest that the kubeconfig would ever be unavailable.

https://developers.linode.com/api/v4/lke-clusters/#post suggests waiting for events.

Even when I've known otherwise, I like to refer to the docs as the source of truth on these matters.

sgmac: Then, let's wait to see what they think, I'm open to suggestions. Thanks for reviewing this.

Charliekenney23:
Hey @sgmac, thanks for starting the discussion and helping to flesh out the foundation for LKE support in Terraform!

I'm closing this as #139 has been merged to account for waiting for the cluster to be ready.

PR #126, opened 2020-02-18
Configure Retries for 429

phillc:
Eventually this should become pkg/retries and most functions become exportable.

### DIFF
--- a/client.go
+++ b/client.go
@@ -158,7 +158,7 @@ func (c *Client) SetToken(token string) *Client {
 	return c
 }
 
-// SetRetries adds retry conditions for Linode Busy. errors and 429s.efault to the given duration.
+// SetRetries adds retry conditions for "Linode Busy." errors and 429s.
 func (c *Client) SetRetries() *Client {
 	addRetryConditional(linodeBusyRetryCondition)
 	addRetryConditional(tooManyRequestsRetryCondition)
--- a/client.go
+++ b/client.go
@@ -158,29 +158,11 @@ func (c *Client) SetToken(token string) *Client {
 	return c
 }
 
-// SetRetries adds retry conditions for Linode Busy. errors and 429s. It also configures resty to
-// lock until enough time has passed to retry the request as determined by the Retry-After response header.
-// If the Retry-After header is not set, SetRetryAfterRateLimit will default to the given duration.
+// SetRetries adds retry conditions for Linode Busy. errors and 429s.efault to the given duration.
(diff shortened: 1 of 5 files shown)


client.go
```diff
@@ -158,24 +158,11 @@ func (c *Client) SetToken(token string) *Client {
...
-				linodeBusy := ok && apiError.Error() == "Linode busy."
-				retry := r.StatusCode() == http.StatusBadRequest && linodeBusy
-				if retry {
-					log.Printf("[INFO] Received error %s - Retrying", apiError)
-				}
-				return retry
-			},
-		)
-
+// SetRetries adds retry conditions for Linode Busy. errors and 429s.efault to the given duration.
```
Charliekenney23: Nit: typo here

retries.go
```diff
@@ -0,0 +1,68 @@
...
+type RetryConditional resty.RetryConditionFunc
+
+var retryConditionals []RetryConditional
+
+// Configures resty to
+// lock until enough time has passed to retry the request as determined by the Retry-After response header.
+// If the Retry-After header is not set, we fall back to value of SetPollDelay.
+func configureRestyRetries(resty *resty.Client) {
+	resty.
+		SetRetryCount(1000).
```
Charliekenney23: I think to default to these loose constraints is fine for now. We should probably circle back and provide a way for consumers to customize the calls to `SetRetryCount` and `SetretryMaxWaitTime`, and merge with defaults.

client.go
```diff
@@ -158,24 +158,11 @@ func (c *Client) SetToken(token string) *Client {
...
+func (c *Client) SetRetries() *Client {
+	addRetryConditional(linodeBusyRetryCondition)
+	addRetryConditional(tooManyRequestsRetryCondition)
```
Charliekenney23: Perhaps these 2 retry condition predicates should be added in `init`? As it stands now, these will be duplicated every time a `linodego.New()` is called.

retries.go
Charliekenney23: I just reread your description and you covered this. My bad!

client.go
phillc: good catch. Maybe the array belongs on the client struct

jnschaeffer: Personally I think these would be good as options in the constructor, but we don't really have a notion of a config object that could be passed in currently.

phillc: If designed from the ground up, perhaps, but most configuration today is handled by method calls on the client

PR #127, opened 2020-02-20
pin golangci-lint

Charliekenney23:


.golangci.yml
```diff
@@ -43,5 +43,5 @@ linters:
     - stylecheck
     - wsl
     - interfacer
+    - gomnd
```
Charliekenney23: Disabling the magic number detector, it just leads to overly verbose code. We should use our own discretion.

Makefile
```diff
@@ -25,11 +32,10 @@ vet:
 	go vet ./...
 
 lint:
-ifndef $(shell command -v $(GOLANGCILINT) -v dot 2> /dev/null)
-	@echo Warning golangci-lint not installed. Skipping lint step
-	@echo Installation: https://github.com/golangci/golangci-lint#install
+ifeq ($(SKIP_LINT), 1)
```
Charliekenney23: Running `SKIP_LINT=1 make` will skip the linting stage.

PR #128, opened 2020-03-02
Move changelog management to github

phillc:


displague:
@phillc You could probably nuke this too: https://github.com/linode/linodego/tree/master/.chglog

phillc:
@displague good call ty

PR #134, opened 2020-03-28
Add SecondaryEntity to Event

stvnjacobs:
For my application, I need access to these values.

I looked at the test fixtures, and these fields are already present in the expected output, so no updates needed there. If you would like testing related to these particular fields, please let me know.


phillc:
Are you sure `Created` is not already there? https://github.com/linode/linodego/blob/master/account_events.go#L183

PR #136, opened 2020-04-11
fix ICMP protocol value

Charliekenney23:


phillc:
If found via a bug, perhaps it needs a test to be filled in.

Charliekenney23:
> If found via a bug, perhaps it needs a test to be filled in.

I will cover this in my next PR coming shortly for implementing `client.GetFirewall`

PR #138, opened 2020-04-12
add GetFirewall method

Charliekenney23:


firewall_rules.go
```diff
@@ -18,7 +18,7 @@ type NetworkAddresses struct {
 
 // A FirewallRule is a whitelist of ports, protocols, and addresses for which traffic should be allowed.
 type FirewallRule struct {
-	Ports     string           `json:"ports"`
+	Ports     string           `json:"ports,omitempty"`
```
Charliekenney23, line 21: This is important, as with ICMP, `ports` is invalid if present in the request.

test/integration/firewalls_test.go
```diff
@@ -19,7 +20,7 @@ var (
 func TestListFirewalls(t *testing.T) {
 	client, _, teardown, err := setupFirewall(t, []firewallModifier{
 		func(createOpts *linodego.FirewallCreateOptions) {
-			createOpts.Label = randString(12, lowerBytes, digits) + "-linodego-testing"
+			createOpts.Label = randString(12, lowerBytes, upperBytes) + "-linodego-testing"
```
Charliekenney23, line 23: A Firewall's label cannot begin with a number, rerunning this threw an error because of it.

PR #139, opened 2020-04-21
implement WaitForLKEClusterReady

Charliekenney23:
`Ready` constitutes a cluster with a healthy api-server and at least one node with the condition `NodeReady=True`.

Waits for:
1. Getting the actual Kubeconfig for building a client from the Linode API (this is usually a minute after the cluster is provisioned).
2. Waiting for a Node to come online and have the `NodeReady` condition (this can take 3-4 minutes).

My 5 minute timeout for the test came from [here](https://github.com/kubernetes/kubernetes/blob/master/test/e2e/framework/util.go#L125).


waitfor.go
```diff
@@ -187,6 +194,91 @@ func (client Client) WaitForLKEClusterStatus(ctx context.Context, clusterID int,
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to build k8s client from LKE cluster kubeconfig: %s", err)
+	}
+	return clientset, nil
+}
+
+// WaitForLKEClusterReady polls with a given timeout for the LKE Cluster's api-server
+// to be healthy and for the cluster to have at least one node with the NodeReady
+// condition true.
+func (client Client) WaitForLKEClusterReady(ctx context.Context, clusterID, timeoutSeconds int) error {
```
displague, line 230: @Charliekenney23, this function will improve the quality of life for LKE developers :+1: 

A slight refactoring could allow you to add extensibility here and reduce the cyclomatic complexity:

```go
package condition
func HasReadyNode(context.Context, *kubernetes.ClientSet) bool {
   // bottom half of your function
}
```
```go
type LKEClientCondition func(context.Context, *kubernetes.ClientSet) bool

// WaitForLKEClientConditions waits for all LKE client conditions to succeed within the timeout period
func WaitForLKEClientConditions(ctx, clusterID, timeoutSeconds, ...LKEClientCondition) error {
  // top half of your function
}
```

This pattern could be used with bespoke criteria, for example, requiring all nodes to be ready or to await a particular deployment to be ready. The typical use, as originally intended, would be:

```go
client.WaitForLKEClientConditions(ctx, lkeClusterID, timeout, conditions.HasReadyNode)
```

What do you think?

asauber: I like this idea. `WaitForLKEClusterReady` could also be provided which calls this as a convenience.

(I was going to say pre-curried, but requiring WaitForLKEClusterReady to take the "rest" of the `ClientConditions` would actually be less convenient than the current signature)

displague: This would keep the Kubernetes client requirements isolated from the rest of the linodego package.

displague: @asauber Sure, `WaitForLKEClusterReady` would look like:

```go
func (client Client) WaitForLKEClusterReady(ctx context.Context, lkeClusterID, timeoutSeconds int) {|
    return client.WaitForLKEClientConditions(ctx, lkeClusterID, timeoutSeconds, condition.HasReadyNode)
}
```

Charliekenney23: Thanks for the idea!

Just pushed [`client.WaitForLKEClusterConditions`](https://github.com/linode/linodego/pull/139/commits/f96d2ee34d93d206cca402184a0d69da48c73a5f#diff-18dc188a67872bccb91f231d01326eecR206) and [`condition.ClusterHasReadyNode`](https://github.com/linode/linodego/pull/139/commits/f96d2ee34d93d206cca402184a0d69da48c73a5f#diff-851683980f0bc25d1450163701749a16R19). Also found a way to record api-server requests (had to change the signature to include [`linodego.LKEClusterPollOptions`](https://github.com/linode/linodego/pull/139/commits/f96d2ee34d93d206cca402184a0d69da48c73a5f#diff-18dc188a67872bccb91f231d01326eecR195) to allow [hooking into the transport](https://github.com/linode/linodego/pull/139/commits/f96d2ee34d93d206cca402184a0d69da48c73a5f#diff-18dc188a67872bccb91f231d01326eecR227)).

lke_clusters.go
```diff
@@ -35,7 +35,7 @@ type LKEClusterCreateOptions struct {
 	NodePools []LKEClusterPoolCreateOptions `json:"node_pools"`
 	Label     string                        `json:"label"`
 	Region    string                        `json:"region"`
-	Version   string                        `json:"version"`
+	Version   string                        `json:"k8s_version"`
```
displague, line 38: `Version` may be surprising to anyone using the Linode API docs as a reference.  I'm sure there are other patterned examples of the go field not matching the upstream name, but this one jumps out. It is understandable because of the change in the API, but I wonder if the field should follow the API here.

waitfor.go
```diff
@@ -187,6 +192,104 @@ func (client Client) WaitForLKEClusterStatus(ctx context.Context, clusterID int,
...
+	ticker := time.NewTicker(client.millisecondsPerPoll * time.Millisecond)
+	defer ticker.Stop()
+
+	var prevLog string
+	var clientset kubernetes.Clientset
+
+	clientReady := func(ctx context.Context, c kubernetes.Clientset) (bool, error) {
+		var err error
+		if clientset == nil {
+			clientset, err = getLKEClusterClientset(ctx, &client, clusterID, options.TransportWrapper)
```
asauber, line 251: In my opinion each of these condition functions should use the `c` which is passed to them, not construct any new clientsets. 

@Charliekenney23 do you think it's strange that the first condition constructs the clientset for all of the other conditions, by abusing a closure?

PR #154, opened 2020-06-23
Add more retry customization

Charliekenney23:
This change
- Moves `Error`, `APIError`, and `APIErrorReason` to `pkg/errors`.
- Adds `errors.New` function in place of `linodego.NewError`.
- Adds `errors.FromString`, `errors.FromError`, and `errors.FromStringer` in place of `linodego.ErrorFromString`, `linodego.ErrorFromError` and `linodego.ErrorFromStringer`.
- Adds a `(Client).SetRetryConditionals([]retry.ConditionFunc)` method in place of `(Client).SetRetries()`.
- Exposes retry conditions `retry.LinodeBusyRetryCondition`, `retry.TooManyRequestsRetryCondition`, and `retry.ServiceUnavailableRetryCondition`.


pkg/errors/errors.go
```diff
@@ -1,31 +1,37 @@
...
 
 const (
-	// ErrorFromString is the Code identifying Errors created by string types
-	ErrorFromString = 1
-	// ErrorFromError is the Code identifying Errors created by error types
-	ErrorFromError = 2
-	// ErrorFromStringer is the Code identifying Errors created by fmt.Stringer types
-	ErrorFromStringer = 3
+	// FromString is the Code identifying Errors created by string types
+	FromString = iota + 1
```
phillc, line 13: This change will need to be called out specifically, as everything else, people will only need to add `errors.` as a prefix, but this one is also a rename.

pkg/errors/errors.go
```diff
@@ -1,31 +1,37 @@
...
 	Message  string
 }
 
-// APIErrorReason is an individual invalid request message returned by the Linode API
+func (e Error) Error() string {
+	return fmt.Sprintf("[%03d] %s", e.Code, e.Message)
+}
+
+// Error implements error.
+var _ error = (*Error)(nil)
```
phillc, line 32: Is this like a compile time assertion/test?

pkg/retry/retry.go
```diff
@@ -0,0 +1,69 @@
...
+
+const retryAfterHeaderName = "Retry-After"
+
+// ConditionFunc represents a function that evaluates whether a request
+// should be retried, given the response and error.
+type ConditionFunc resty.RetryConditionFunc
+
+// ComposeConditional composes a ConditionFunc function from an array
+// of ConditionFunc functions.
+func ComposeConditionFunc(conditionals ...ConditionFunc) ConditionFunc {
```
phillc, line 21: Did you intend to use this instead of the retry conditional above? Your intent I assume is that you could technically use a single retry condition without composing?

pkg/errors/errors.go
Charliekenney23: Oops yeah, I dunno why I left this in-- I'll remove it.

But, FWIW, https://golang.org/doc/effective_go.html#blank_implements.

pkg/retry/retry.go
Charliekenney23: Should also remove this, my initial intent was to use this in `(client).configureRetries` to step through `c.retryConditionals`, but I ended up not doing that so that I could keep our log before retrying.

pkg/errors/errors.go
Charliekenney23: When it comes time to cut the release, I'll be referencing what I called out in the above description.

pkg/errors/errors.go
phillc: I thought it was fine, was just curious.

PR #158, opened 2020-07-27
Add LKEClusterPoolDisk type and disks field to LKEClusterPool

jnschaeffer:
### General:
`disks` was missing from `LKEClusterPool`. I think this fixes that.


phillc:
Maybe a quick test case to prove it? https://github.com/linode/linodego/blob/master/test/integration/lke_cluster_pools_test.go

PR #161, opened 2020-08-27
Remove k8s dependencies from core module

### LINKED ISSUE #155: Remove dependency on k8s.io/client-go
## General:

* [x] Have you removed all sensitive information, including but not limited to access keys and passwords?
* [x] Have you checked to ensure there aren't other open or closed [Pull Requests](../../pulls) for the same bug/feature/question?

----

We would like to see linodego remove its dependency on `k8s.io/client-go` package here: https://github.com/linode/linodego/tree/master/internal/kubernetes

This causes problem as users may depend on a different version of k8s.io/client-go. And k8s.io/client-go breaks its compatibility every 3 months.

Proposed Solution:
Please expose that package as its own Go module. You can do that keeping that package in this git repo or in separate repos.

phillc:
I have a big changelog entry to write, as things have moved around.

Resolves #155


phillc:
Added:

https://github.com/linode/linodego/k8s module, which is dependent on kubernetes modules.

Changed:

Kubernetes modules are no longer a dependency of the base linodego module.

`WaitForLKEClusterReady` has moved from https://github.com/linode/linodego to https://github.com/linode/linodego/k8s/pkg/condition and now takes the linodego client as an argument.

`ClusterHasReadyNode` has moved from https://github.com/linode/linodego/pkg/condition to https://github.com/linode/linodego/k8s/pkg/condition

`ClusterConditionFunc` now takes an options struct instead of a kubernetes client.

`test` is now its own module, allowing integration tests to be isolated, and its kubernetes dependencies not part of the base module.

k8s/pkg/condition/lke.go
```diff
@@ -0,0 +1,43 @@
...
+
+	for _, node := range nodes.Items {
+		for _, condition := range node.Status.Conditions {
+			if condition.Type == corev1.NodeReady && condition.Status == corev1.ConditionTrue {
+				return true, nil
+			}
+		}
+	}
+
+	return false, errors.New("no nodes in cluster are ready")
```
jnschaeffer, line 35: Does it make sense to return an error if no nodes are ready? In general, I would expect a function like this to return `false` if the condition successfully evaluated to false, then for the upstream function using it to fail eventually on timeout.

phillc: You make a point. I'll defer to @Charliekenney23 , but this PR is mostly just a refactor without functional change, so should probably be part of different PR

Charliekenney23: John is right, it definitely doesn't match the k8s waiter convention. But yes, this is old code I added initially, so I can do this in a subsequent PR before we release.

PR #165, opened 2020-09-28
make ClusterConditionFunc's adhere to k8s waiter convention

Charliekenney23:
See: https://godoc.org/gopkg.in/kubernetes/client-go.v1/1.5/pkg/util/wait#ConditionFunc
> ConditionFunc returns true if the condition is satisfied, or an error if the loop should be aborted.


phillc (approved):
needs noted as backwards incompat

PR #173, opened 2020-12-04
add object-acl get and update methods

Charliekenney23:


object_storage_object.go
```diff
@@ -0,0 +1,83 @@
...
+}
+
+type ObjectStorageObjectURL struct {
+	URL    string `json:"url"`
+	Exists bool   `json:"exists"`
+}
+
+type ObjectStorageObjectACLConfig struct {
+	ACL    string `json:"acl"`
+	ACLXML string `json:"acl_xml"`
```
phillc, line 24: Should probably have a test exert the xml version

phillc (changes requested):

object_storage_object.go
Charliekenney23: So my original interpretation of the "custom" ACL was wrong. You can specify "custom", but you cannot include XML in a PUT request. The `acl_xml` field is only present in the responses from the API.

phillc:
Why bother validating?

phillc (approved):

PR #185, opened 2021-03-15
Add ACL and CorsEnabled options to Object Storage

LBGarber:
- Add `ACL` and `ObjectStorageACL` enums for configuring the access control of a new Object Storage bucket.
- Add `CorsEnabled` option for bucket creation.

### DIFF
--- a/object_storage_buckets.go
+++ b/object_storage_buckets.go
@@ -42,7 +42,21 @@ func (i *ObjectStorageBucket) UnmarshalJSON(b []byte) error {
 type ObjectStorageBucketCreateOptions struct {
 	Cluster string `json:"cluster"`
 	Label   string `json:"label"`
+
+	ACL         ObjectStorageACL `json:"acl,omitempty"`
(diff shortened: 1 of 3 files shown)


Charliekenney23 (approved):

object_storage_buckets.go
```diff
@@ -42,8 +42,22 @@ func (i *ObjectStorageBucket) UnmarshalJSON(b []byte) error {
 type ObjectStorageBucketCreateOptions struct {
 	Cluster string `json:"cluster"`
 	Label   string `json:"label"`
+
+	ACL         ObjectStorageACL `json:"acl,omitempty"`
+	CorsEnabled *bool            `json:"cors_enabled,omitempty"`
```
phillc, line 47: pointer bools are always awkward, but I guess necessary here

phillc (approved):

object_storage_buckets.go
LBGarber: Yeah, it seems like this is the best option unless we abstracted bools in some way.

PR #186, opened 2021-03-16
Add Object Storage bucket access update and get methods

LBGarber:
- Add `GetObjectStorageBucketAccess(...)` and `UpdateObjectStorageBucketAccess(...)`
- Add tests `TestGetObjectStorageBucketAccess` and `TestUpdateObjectStorageBucketAccess`

### DIFF
--- a/test/integration/object_storage_buckets_test.go
+++ b/test/integration/object_storage_buckets_test.go
@@ -10,12 +10,14 @@ import (
 )
 
 var testObjectStorageBucketCreateOpts = ObjectStorageBucketCreateOptions{
-	Cluster:     "us-east-1",
-	Label:       fmt.Sprintf("linodego-test-bucket-%d", time.Now().UnixNano()),
+	Cluster: "us-east-1",
+	Label:   fmt.Sprintf("linodego-test-bucket-%d", time.Now().UnixNano()),
 }
 
 func TestCreateObjectStorageBucket(t *testing.T) {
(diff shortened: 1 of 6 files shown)


test/integration/object_storage_buckets_test.go
```diff
@@ -93,6 +93,65 @@ func TestListObjectStorageBuckets(t *testing.T) {
 	}
 }
 
+func TestGetObjectStorageBucketAccess(t *testing.T) {
+	corsEnabled := false
+
+	// Inject ACL and Cors config into create opts
+	oldCreateOpts := testObjectStorageBucketCreateOpts
+	defer func() { testObjectStorageBucketCreateOpts = oldCreateOpts }()
```
Charliekenney23, line 101: I would just construct a new `createOpts` here instead, it would be cleaner. Also, each test runs in a separate goroutine so all tests that reference the `testObjectStorageBucketCreateOpts` could have your modified options.

Charliekenney23 (changes requested):

test/integration/object_storage_buckets_test.go
LBGarber: Yeah, that felt kind of hacky. I guess we might need to adapt `setupObjectStorageBucket` unless there's a better way to do it that I'm missing.

Charliekenney23 (approved):

Charliekenney23 (approved):

PR #187, opened 2021-03-24
decouple SetRetryWaitTime and SetPollDelay

Charliekenney23:
SetPollDelay should not set the minimum retry wait time.


LBGarber (approved):
Looks good!

PR #190, opened 2021-04-13
run tests in github actions

Charliekenney23:
Removes our old TravisCI configuration and runs linter/tests on every push.


LBGarber (approved):
Looks good!

PR #198, opened 2021-05-12
use runtime/debug to source linodego Version

Charliekenney23:
This change exposes the package version dynamically via `linodego.Version` and in the user agent.


LBGarber (approved):
Looks good!

PR #203, opened 2021-06-01
Add Domain zone-file endpoint

LBGarber:
- Fix existing test suite for Domains and Domain Records
- Add support for Domain zone-file endpoint

### DIFF
--- a/domains.go
+++ b/domains.go
@@ -52,6 +52,11 @@ type Domain struct {
 	TTLSec int `json:"ttl_sec"`
 }
 
+// DomainZoneFile represents the Zone File of a Domain
+type DomainZoneFile struct {
+	ZoneFile []string `json:"zone_file"`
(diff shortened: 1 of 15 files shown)


displague:
I'm curious if the zone response may be a 200 with a null file under some conditions, such as a zone file that failed to generate (or has not been generated yet, as in the case of new domains).

https://registry.terraform.io/providers/linode/linode/latest/docs/resources/domain#status shows that the status may be something other than active/disabled.  The API documentation says that this field is either `active` or `disabled`, so what is expressed in TF docs may no longer be the case: https://www.linode.com/docs/api/domains/#domain-view__responses.

Charliekenney23 (approved):

Charliekenney23:
@displague The API will always send the zone file regardless of status

PR #206, opened 2021-06-22
improve generic html error propagation

Charliekenney23:
When our API LBs fail to reach a server upstream they sometimes return generic HTML error responses. This change improves error propagation in logs for this case by outputting the entire body of the response as well as the status code.

In the event that an API NB throws a 500:

```
[001] Unexpected Content-Type: Expected: application/json, Received: text/html
```

now looks like:

```
[500] Unexpected Content-Type: Expected: application/json, Received: text/html
Response body: <html>
<head><title>500 Internal Server Error</title></head>
<body bgcolor="white">
<center><h1>500 Internal Server Error</h1></center>
```

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -63,12 +63,13 @@ func coupleAPIErrors(r *resty.Response, err error) (*resty.Response, error) {
 
 		if responseContentType != expectedContentType {
 			msg := fmt.Sprintf(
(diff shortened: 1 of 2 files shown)


LBGarber (approved):
Perfect!

PR #210, opened 2021-07-22
WaitForEventFinished respects minStart and action

LBGarber:
This change defers event filtering logic for creation date and action to the Linode API, which should reduce the number of events returned by the API.

### DIFF
--- a/waitfor.go
+++ b/waitfor.go
@@ -261,13 +261,12 @@ func (client Client) WaitForEventFinished(ctx context.Context, id interface{}, e
 	titledEntityType := strings.Title(string(entityType))
 	filterStruct := map[string]interface{}{
 		// Nor is action
-		//"action": action,
+		"action": action,
 
-		// Created is not correctly filtered by the API
-		// We'll have to verify these values manually, for now.
-		//"created": map[string]interface{}{
-		//	"+gte": minStart.Format(time.RFC3339),
-		//},
+		"created": map[string]interface{}{
+			// The API uses UTC time, so we need to ensure the time is converted
+			"+gte": minStart.UTC().Format("2006-01-02T15:04:05"),
+		},
 
 		// With potentially 1000+ events coming back, we should filter on something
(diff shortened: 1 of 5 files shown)


waitfor.go
```diff
@@ -261,13 +261,13 @@ func (client Client) WaitForEventFinished(ctx context.Context, id interface{}, e
...
 
-		// Created is not correctly filtered by the API
-		// We'll have to verify these values manually, for now.
-		//"created": map[string]interface{}{
-		//	"+gte": minStart.Format(time.RFC3339),
-		//},
+		"created": map[string]interface{}{
+			// A literal date format is necessary because events
+			// cannot be filtered on `gte created` with a full RFC 3339 date.
+			"+gte": minStart.Format("2006-01-02T15:04:05"),
```
displague, line 269: Without RFC3339 the timezone will be omitted.  I assume this will be safe regardless of the timezone where this is being used. (minStart will have a timezone baked into it, or will default to UTC, which is the timezone the API uses in responses).

https://github.com/linode/linodego/pull/210/files#diff-904bd3d16d52b4d495ecbb5db8ac1d4db55ce10360f91987e9dece5dfa2790c0R376-R384 may also need revisiting since the filtering will be server-side rather than client-side.

displague: https://github.com/linode/linodego/pull/210/files#diff-904bd3d16d52b4d495ecbb5db8ac1d4db55ce10360f91987e9dece5dfa2790c0R341-R344 also.

It's probably safe to keep the local checks as sanity checks, but also probably not needed if the API is doing the right thing 🤷

displague: @Dorthu suspects this is fine, fwiw

Dorthu: It's true, Marques reached out and i verified that filtering events by "created" and "action" works as expected.

LBGarber: Perfect, thanks Will!

Charliekenney23 (approved):

PR #215, opened 2021-09-01
Remove 'seen' filter from WaitForEventFinished

LBGarber:
This change removes the `seen` filter from the API requests made during `WaitForEventFinished`. This flag should no longer be necessary as we can now use API-level timestamp filtering to narrow down event results.

See https://github.com/linode/terraform-provider-linode/issues/454


jriddle-linode (approved):
LGTM

PR #216, opened 2021-09-10
fix: don't retry requests during API maintenance events

tjhop:
During Linode API maintenance events (like the [upcoming maintenance on 09/23/21](https://status.linode.com/incidents/khfqpfsfdh0g)), the API will return a `503` and set a header (`X-MAINTENANCE-MODE`) indicating that maintenance is in progress. The linodego client shouldn't bother retrying requests during maintenance events, as it's known and expected that it will not work. 

An additional test for maintenance mode has been added:

```
=== RUN   TestLinodeServiceMaintenanceModeRetryCondition
2021/09/10 10:33:34 [INFO] Linode API is under maintenance, request will not be retried
--- PASS: TestLinodeServiceMaintenanceModeRetryCondition (0.00s)
```

### DIFF
--- a/retries.go
+++ b/retries.go
@@ -57,19 +57,17 @@ func tooManyRequestsRetryCondition(r *resty.Response, _ error) bool {
 }
 
 func serviceUnavailableRetryCondition(r *resty.Response, _ error) bool {
-	var retry bool
-	if r.StatusCode() == http.StatusServiceUnavailable {
-		// During maintenance events, the API will return a 503 and add
-		// an `X-MAINTENANCE-MODE` header. Don't rety during maintenance
-		// events, only for legitimate 503s.
-		if r.Header().Get(maintenanceModeHeaderName) != "" {
-			log.Printf("[INFO] Linode API is under maintenance, request will not be retried")
-		} else {
-			retry = true
-		}
+	serviceUnavailable := r.StatusCode() == http.StatusServiceUnavailable
+
+	// During maintenance events, the API will return a 503 and add
+	// an `X-MAINTENANCE-MODE` header. Don't retry during maintenance
+	// events, only for legitimate 503s.
+	if serviceUnavailable && r.Header().Get(maintenanceModeHeaderName) != "" {
+		log.Printf("[INFO] Linode API is under maintenance, request will not be retried")
+		return false
 	}
 
-	return retry
+	return serviceUnavailable
 }
 
 func requestTimeoutRetryCondition(r *resty.Response, _ error) bool {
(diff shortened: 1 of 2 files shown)


retries.go
```diff
@@ -54,7 +57,19 @@ func tooManyRequestsRetryCondition(r *resty.Response, _ error) bool {
...
 
 func serviceUnavailableRetryCondition(r *resty.Response, _ error) bool {
-	return r.StatusCode() == http.StatusServiceUnavailable
+	var retry bool
+	if r.StatusCode() == http.StatusServiceUnavailable {
+		// During maintenance events, the API will return a 503 and add
+		// an `X-MAINTENANCE-MODE` header. Don't rety during maintenance
+		// events, only for legitimate 503s.
+		if r.Header().Get(maintenanceModeHeaderName) != "" {
+			log.Printf("[INFO] Linode API is under maintenance, request will not be retried")
```
tjhop, line 66: Should we also log a message telling users where they can go for more information? Ie, add something like

> Please see status.linode.com for more details

🤔

retries.go
```diff
@@ -54,7 +57,19 @@ func tooManyRequestsRetryCondition(r *resty.Response, _ error) bool {
...
+		} else {
+			retry = true
+		}
+	}
+
+	return retry
```
LBGarber, line 72: We might be able to simplify the logic here a bit :+1: 

```suggestion
	serviceUnavailable := r.StatusCode() == http.StatusServiceUnavailable

	// During maintenance events, the API will return a 503 and add
	// an `X-MAINTENANCE-MODE` header. Don't retry during maintenance
	// events, only for legitimate 503s.
	if serviceUnavailable && r.Header().Get(maintenanceModeHeaderName) != "" {
		log.Printf("[INFO] Linode API is under maintenance, request will not be retried")
		return false
	}

	return serviceUnavailable
```

LBGarber (changes requested):
Awesome PR! Just one small stylistic suggestion and we should be good to go!

retries.go
LBGarber: I like this idea a lot! Feel free to add it

retries.go
tjhop: Fine by me! I don't mind switching to inline conditionals, they're more prevalent in this bit of code anyway. 👍

retries.go
tjhop: Added in 6d36a3d069c72b413f6741fdf26fbadfc3245065 👍

LBGarber (approved):
Looks good :+1:

PR #219, opened 2021-09-20
Rework comparison to handle order/order_by and filters with no logicaloperators

jriddle-linode:


comparison.go
```diff
@@ -82,34 +42,48 @@ func (f *Filter) JSON() string {
...
 	Value    interface{}
 }
 
-func (c *Comparison) GetChildren() []FilterNode {
+func (c *Comp) GetChildren() []FilterNode {
 	return []FilterNode{}
 }
 
-func (c *Comparison) JSON() string {
+func (c *Comp) JSON() string {
```
displague, line 71: There is an opportunity to use MarshalJSON to get some of this for free and have a reusable interface (rather than exporting a new JSON function that doesn't emit actual JSON).

https://pkg.go.dev/encoding/json

LBGarber (approved):
Looks good!

PR #225, opened 2021-11-10
Make SetAPIVersion respect previous calls

LBGarber:
This pull request resolves an issue where calling `SetAPIVersion` multiple times would result in a broken resulting `BaseURL`.


jriddle-linode (approved):
LGTM

PR #243, opened 2022-04-07
adding a public function to add retry conditions to Resty

jriddle-linode:


LBGarber (approved):
Looks good! I'm glad we're isolating the public logic from the private logic to avoid any implicit weirdness.

PR #244, opened 2022-04-07
add retry bool to waitfork8savailable

jriddle-linode:


waitfor.go
```diff
@@ -237,7 +240,9 @@ func (client Client) WaitForLKEClusterConditions(
 			case <-ticker.C:
 				result, err := condition(ctx, conditionOptions)
 				if err != nil {
-					return err
+					if !options.Retry {
+						return err
+					}
```
LBGarber, line 245: We might want to add a log statement here so we can more easily debug issues that may be retried on.

```suggestion
					}

					log.Printf("[WARN] Ignoring WaitForLKEClusterConditions conditional error %s", err)
```

LBGarber (changes requested):
Looks good! Just one small thing

PR #253, opened 2022-05-27
Updating linodego to go 1.18

jriddle-linode:


LBGarber:
The `TestDatabase_Suite` fixtures seem to be broken; could you rerun them?

PR #268, opened 2022-08-05
Pagination Rewrite

jriddle-linode:


test/integration/mysql_test.go
```diff
@@ -30,6 +30,7 @@ var testMySQLCreateOpts = linodego.MySQLCreateOptions{
 }
 
 func TestDatabase_MySQL_Suite(t *testing.T) {
+	t.Skip("dbaas is currently having issues provisioning")
```
LBGarber, line 33: Could we drop these skips before we merge?

instances.go
```diff
@@ -74,7 +75,7 @@ type InstanceAlert struct {
 // InstanceBackup represents backup settings for an instance
 type InstanceBackup struct {
 	Enabled  bool `json:"enabled,omitempty"`
-	Schedule struct {
+	Schedule *struct {
```
LBGarber, line 78: This is _technically_ a breaking chance but it seems like it never worked correctly in the first place. We should include this in our release notes just in case.

jriddle-linode: doesn't actually change anything sadly, removing it

LBGarber (approved):

PR #278, opened 2022-10-10
new: Add caching framework for static endpoint responses

LBGarber:
This change adds a framework for automatically caching static endpoint responses. This should significantly reduce the API load and improve execution time issues originating from duplicate requests.


client.go
```diff
@@ -189,6 +205,127 @@ func (c *Client) addRetryConditional(retryConditional RetryConditional) *Client
...
+
+	entry, ok := c.cachedEntries[u.Path]
+	if !ok {
+		return nil, nil
+	}
+
+	// Handle expired entries
+	elapsedTime := time.Since(entry.Created)
+
+	shouldExpire := elapsedTime > c.cacheExpiration
```
jcallahan-akamai, line 274: Would `hasExpired` be a better variable name?

jriddle-linode (changes requested):
Everything looks great can we add some documentation for using cache?

jriddle-linode (approved):

PR #286, opened 2022-12-21
retry calls on http2 goaway frames

### LINKED ISSUE #283: [Feature]: Retry on HTTP2 GOAWAY frames
### Description

Occasionally, we'll see errors like this when using Linode-based service discovery in Prometheus:
```console
ts=2022-12-01T15:21:15.357Z caller=refresh.go:99 level=error component="discovery manager scrape" discovery=linode msg="Unable to refresh target groups" err="[002] http2: server sent GOAWAY and closed the connection; LastStreamID=1999, ErrCode=NO_ERR
Dec 01 10:21:15 prometheus2 prometheus[2268065]: 2022/12/01 10:21:15.357289 ERROR RESTY http2: server sent GOAWAY and closed the connection; LastStreamID=1999, ErrCode=NO_ERROR, debug="", Attempt 1
```

What happens is that the connection to api.linode.com gets reset because we hit the maximum number of requests for a persistent connection (e.g. [nginx defaults to 1000](https://nginx.org/en/docs/http/ngx_http_v2_module.html#http2_max_requests)) and the connection is closed by the server after it sends a [GOAWAY](https://httpwg.org/specs/rfc7540.html#GOAWAY). We should be able to safely retry these by opening a new connection.

There's likely more nuance to it, though. As I believe you need to validate the accompanying error with the GOAWAY frame and check on the last stream ID rather than just naively retrying every time.

Some helpful resources I've found are:
- https://github.com/golang/go/issues/18639
- [Pagerduty's retry logic](https://github.com/PagerDuty/go-pdagent/blob/8d1c23d752f256deb02d2a790b89768a71d67e7d/pkg/common/retrytransport.go#L118-L138)
- [Azure AKS retry logic](https://github.com/Azure/aks-deployer/blob/a1c22f67e96315b9eeb3937558a90925dc5f7a3e/pkg/retry/error.go)
- [net/http2 GoAwayError type](https://github.com/golang/net/blob/ecf7fda6a59edbedc58c1207f55285a9ad98e10a/http2/transport.go#L2132-L2143)
- [Stackoverflow, of course](https://stackoverflow.com/questions/45209168/http2-server-sent-goaway-and-closed-the-connection-laststreamid-1999?rq=1)

### Example Code

```go
// retries.go
func requestGOAWAYRetryCondition(_ *resty.Response, e error) bool {
	return errors.As(e, &http2.GoAwayError{})
}
```

jriddle-linode:
## 📝 Description

whenever resty receives a goaway frame retry.

## ✔️ How to Test

`make ARGS='-run TestClient_GoAwayRetry' fixtures`

resolves #283

### DIFF
--- a/.github/workflows/gosec_pr.yml
+++ b/.github/workflows/gosec_pr.yml
@@ -1,16 +1,15 @@
 name: Gosec Scan
 on:
-  pull_request:
-    types: [opened, reopened]
+  pull_request: null
 
 jobs:
-  tests:
+  gosec_scan:
     runs-on: ubuntu-latest
     env:
       GO111MODULE: on
     steps:
       - name: Checkout Source
-        uses: actions/checkout@v2
+        uses: actions/checkout@5a4ac9002d0be2fb38bd78e4b4dbde5606d7042f # pin@v2
       - name: Run Gosec Security Scanner
         uses: securego/gosec@master
         with:
--- a/.github/workflows/gosec_pr.yml
+++ b/.github/workflows/gosec_pr.yml
@@ -1,0 +1,17 @@
+name: Gosec Scan
+on:
+  pull_request:
+    types: [opened, reopened]
+
+jobs:
+  tests:
+    runs-on: ubuntu-latest
+    env:
+      GO111MODULE: on
+    steps:
+      - name: Checkout Source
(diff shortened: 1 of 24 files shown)


lgarber-akamai (approved):
Nice work!

PR #291, opened 2023-01-18
Add PrivateIP option for cloning Linodes

zliang-akamai:
Linode API v4 now supports having a new linode created from cloning being assigned a private net IP. Adding corresponding feature in linodego.

### DIFF
--- a/test/integration/instances_test.go
+++ b/test/integration/instances_test.go
@@ -376,8 +376,8 @@ func TestInstance_Rebuild(t *testing.T) {
 }
 
 func TestInstance_Clone(t *testing.T) {
-	client, instance, teardown, err := setupInstance(t, "fixtures/TestInstance_Clone")
-	defer teardown()
+	client, instance, teardownOriginalLinode, err := setupInstance(t, "fixtures/TestInstance_Clone")
+	defer teardownOriginalLinode()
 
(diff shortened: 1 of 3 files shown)


jriddle-linode (approved):

test/integration/instances_test.go
```diff
@@ -375,6 +375,54 @@ func TestInstance_Rebuild(t *testing.T) {
...
+		*cloned_instance.Created,
+		240,
+	)
+
+	if err != nil {
+		t.Error(err)
+	}
+
+	// clean up
+	client.DeleteInstance(context.Background(), cloned_instance.ID)
```
lgarber-akamai, line 423: Could we add a step to validate that the cloned instance has a private IP, etc?

lgarber-akamai (changes requested):
Looks great! Just one suggestion

test/integration/instances_test.go
zliang-akamai: Added, thank you!

lgarber-akamai:
/acctest sha=8626ee6a68d707d638f802e17d176335f4fedb26 tests=TestInstance_Clone

lgarber-akamai (approved):

PR #293, opened 2023-01-27
Grants List

zliang-akamai:
<strike>#### Current issues:
1. [The existing utility function](https://github.com/linode/linodego/blob/cfc31f1d4d543943a8b880221304eaa64bd6cf5f/errors.go#L64) that's used to access the API will throw an error when 204 response received because of the expected `content-type` `application/json` defined [here](https://github.com/linode/linodego/blob/cfc31f1d4d543943a8b880221304eaa64bd6cf5f/client.go#L111).
2. There is not an obvious way to test grants list because the test client cannot access the token of a newly created user with some grants. And only that user (or app with the token of the user) can access its own grants via grants list API.

#### Possible and hopefully not the only solutions:
1. Manually testing and merge without an integration test. (Oh no!)
2. Making this ticket as blocked until we fix the utility function. (We will do it, right?)

Update 01/27:
Potential braking change: ignoring content type mismatch error when the response body is empty (204 response).
</strike>
Update 01/31:
Made this PR simple and postpone the potential breaking change.
Grants API docs added database options/attributes recently, so updated it here as well.

### DIFF
--- a/profile_grants_list.go
+++ b/profile_grants_list.go
@@ -1,0 +1,25 @@
+package linodego
+
+import (
+	"context"
+	"encoding/json"
+)
+
+type GrantsListResponse = UserGrants
+
+
+func (c *Client) GrantsList(ctx context.Context) (*GrantsListResponse, error) {
+	e := "profile/grants"
+	r, err := coupleAPIErrors(c.R(ctx).Get(e))
+	if err != nil {
+		return nil, err
+	}
+
+	if r.StatusCode() == 204 {
+		return nil, nil
+	}
+	var result GrantsListResponse
+	err = json.Unmarshal(r.Body(), &result)
+
+	return &result, err
+}
--- a/profile_grants_list.go
+++ b/profile_grants_list.go
@@ -7,7 +7,6 @@ import (
 
 type GrantsListResponse = UserGrants
 
-
 func (c *Client) GrantsList(ctx context.Context) (*GrantsListResponse, error) {
 	e := "profile/grants"
 	r, err := coupleAPIErrors(c.R(ctx).Get(e))
--- a/profile_grants_list.go
+++ b/profile_grants_list.go
@@ -2,6 +2,7 @@ package linodego
 
 import (
 	"context"
+	"log"
 )
 
 type GrantsListResponse = UserGrants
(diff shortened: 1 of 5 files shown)


zliang-akamai:
/acctest 108be8f1706022f0df7a9b61b91a1aa270a9c1a0 tests=TestGrantsList

zliang-akamai:
/acctest 108be8f1706022f0df7a9b61b91a1aa270a9c1a0 tests=TestGrantsList

zliang-akamai:
/acctest 108be8f1706022f0df7a9b61b91a1aa270a9c1a0 tests=TestGrantsList

jriddle-linode (approved):

zliang-akamai:
/acctest 1105a115ddedf0e4837167f1385e968159c8bbb3 tests=TestGrantsList

zliang-akamai:
/acctest 1105a115ddedf0e4837167f1385e968159c8bbb3 tests=TestUserGrants_Update

zliang-akamai:
/acctest 1105a115ddedf0e4837167f1385e968159c8bbb3 tests=TestUserGrants_UpdateNoAccess

profile_grants_list.go
```diff
@@ -0,0 +1,25 @@
...
+
+func (c *Client) GrantsList(ctx context.Context) (*GrantsListResponse, error) {
+	e := "profile/grants"
+	r, err := coupleAPIErrors(c.R(ctx).SetResult(GrantsListResponse{}).Get(e))
+	if err != nil {
+		return nil, err
+	}
+
+	if r.StatusCode() == 204 {
+		log.Printf(
```
lgarber-akamai, line 18: Hmm, wonder if we should add a `SetWarnings(enabled)` function to `Client` to enable and disable warnings like these

lgarber-akamai:
/acctest sha=1105a115ddedf0e4837167f1385e968159c8bbb3 tests=TestGrantsList

profile_grants_list.go
zliang-akamai: Yea, that would be very helpful. Just realized that for this PR, actually, this warning is never reached. I am going to remove it. I will gather the warning/info log as well as 204 handling into another utility function in another PR.

lgarber-akamai (approved):
Nice work!

PR #298, opened 2023-02-14
fix: Repair broken tests after rebase

lgarber-akamai:
## 📝 Description

This change repairs some tests that were broken after a rebase.

## ✔️ How to Test

```make testint```

### DIFF
--- a/test/integration/account_test.go
+++ b/test/integration/account_test.go
@@ -2,7 +2,6 @@ package integration
 
 import (
 	"context"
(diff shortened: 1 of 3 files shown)


zliang-akamai:
Can we place something like `example@example.com` in the email field instead then we won't have to modify the test case?

zliang-akamai (approved):

jriddle-linode (approved):

PR #315, opened 2023-03-24
init

yec-akamai:
## 📝 Description

Update functions to escape labels correctly in urls. 

## ✔️ How to Test

```
make testunit
make testint
```


lgarber-akamai (approved):
Nice work!

PR #316, opened 2023-03-24
Add size,objects fields back to obj bucket

yec-akamai:
## 📝 Description

The `size` and `objects` were removed in 2019 due to a beta change. Now they've been added back to the [object storage bucket API response](https://www.linode.com/docs/api/object-storage/#object-storage-bucket-view). We need to add them back in linodego and terraform (follow-up).

## ✔️ How to Test

```
make testunit
make testint
```

### DIFF
--- a/object_storage_buckets.go
+++ b/object_storage_buckets.go
@@ -17,6 +17,8 @@ type ObjectStorageBucket struct {
 
 	Created  *time.Time `json:"-"`
 	Hostname string     `json:"hostname"`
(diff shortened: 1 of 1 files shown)


jcallahan-akamai (approved):
Looks good. Not required, but we could probably expand the tests to look for these specific fields.

ezilber-akamai (approved):

zliang-akamai (approved):

PR #320, opened 2023-04-19
Fix lint

zliang-akamai:
## 📝 Description

Fix lints, remove deprecated linters

## ✔️ How to Test
`make test`

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -2,6 +2,11 @@ run:
   tests: false
 
 linters-settings:
+  revive:
+    rules:
+      - name: unused-parameter
+        severity: warning
+        disabled: true
   errcheck:
     check-type-assertions: true
     check-blank: true
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -29,6 +29,13 @@ linters-settings:
 linters:
   enable-all: true
   disable:
+    # deprecated linters
+    - deadcode
+    - ifshort
+    - varcheck
+    - nosnakecase
+    ####################
+
     - bodyclose
     - contextcheck
     - nilerr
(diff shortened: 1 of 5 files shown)


yec-akamai (commented):
Can you add some context on why we're disabling these lints? Some of them seem useful to me, i.e. `unused-parameter`

zliang-akamai:
@yec-akamai 
It conflicts with our existing code base. This generally requires us to name a unused parameter as an underscore `_`, which is kind of ugly...

yec-akamai:
> 

Umm, I think it's very common to name unused ones as an underscore. We probably should refactor our code base instead of disabling it.

.golangci.yml
```diff
@@ -2,6 +2,11 @@ run:
   tests: false
 
 linters-settings:
+  revive:
```
yec-akamai, line 5: I look around and there are many examples even in our codebase that follows the rule of using underscore, i.e. https://github.com/linode/linodego/blob/cfc31f1d4d543943a8b880221304eaa64bd6cf5f/instances.go#L279 I'm not seeing any disadvantage of doing it.

yec-akamai (changes requested):

.golangci.yml
zliang-akamai: The difference is this was in function body, and the unused variable in a function should be mark as an underscore which is enforced by the go lang compiler. The linter enforce it for the arguments list. It requires some changes like this:
image

I wouldn't oppose to apply this linter, and I just committed [the fix](https://github.com/linode/linodego/pull/320/commits/2d9cd83e40450f7c52a6c4e87dfd589aa54a0408).

zliang-akamai: Ideally we should remove this parameter but it will be a breaking change if we do so.

yec-akamai (approved):

PR #324, opened 2023-04-25
Simplified Acctest Command Dispatching Workflow

zliang-akamai:
## 📝 Description

When we created the acctest command workflows, GitHub doesn't support using `GITHUB_TOKEN` to trigger another workflow [but later they added some exception](https://github.blog/changelog/2022-09-08-github-actions-use-github_token-with-workflow_dispatch-and-repository_dispatch/).

We used a workaround to generate a token as a GitHub App to trigger an `repository_dispatch` event to trigger the integration test. With the newly added exception, now I believe we can trigger the integration test directly by the `repository_dispatch` event triggered by the built-in `GITHUB_TOKEN`, and we can now remove the action that was used to generate GitHub App token.

## ✔️ How to Test

Will only be testable in this repo after merging.

Success on my forked repo:
https://github.com/zliang-akamai/linodego/pull/2#issuecomment-1522019490

### DIFF
--- a/.github/workflows/acctest_command.yml
+++ b/.github/workflows/acctest_command.yml
@@ -9,20 +9,10 @@ jobs:
     runs-on: ubuntu-latest
     if: ${{ github.event.issue.pull_request }}
     steps:
-      - name: Generate App Installation Token
-        id: generate_token
-        uses: tibdex/github-app-token@v1
-        with:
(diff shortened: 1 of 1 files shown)


yec-akamai (approved):
lgtm

ezilber-akamai (approved):
Looks good!

PR #328, opened 2023-05-16
feat: support query params in list options

wbh1:
## 📝 Description

This will allow for setting custom query parameters in API calls by providing a map.

## ✔️ How to Test

Include QueryParams in your ListOptions variable for endpoints that support query parameters.

N/A

### DIFF
--- a/pagination.go
+++ b/pagination.go
@@ -23,7 +23,7 @@ type PageOptions struct {
 }
 
 // ListOptions are the pagination and filtering (TODO) parameters for endpoints
-//nolint
+// nolint
 type ListOptions struct {
 	*PageOptions
 	PageSize int    `json:"page_size"`
--- a/pagination.go
+++ b/pagination.go
@@ -140,6 +140,13 @@ func flattenQueryStruct(val any) (map[string]string, error) {
 		reflectVal = reflect.Indirect(reflectVal)
 	}
 
+	if reflectVal.Kind() != reflect.Struct {
+		return nil, fmt.Errorf(
+			"expected struct type for the QueryParams but got: %s",
+			reflectVal.Kind().String(),
+		)
+	}
+
 	valType := reflectVal.Type()
 
 	for i := 0; i < valType.NumField(); i++ {
--- a/pagination.go
+++ b/pagination.go
@@ -153,6 +153,16 @@ func flattenQueryStruct(val any) (map[string]string, error) {
 			return nil, fmt.Errorf("invalid query param tag: %s", currentField.Name)
 		}
 
+		// Skip if it's a zero value
+		if valField.IsZero() {
+			continue
+		}
+
+		// Deref the pointer is necessary
+		if valField.Kind() == reflect.Ptr {
+			valField = reflect.Indirect(valField)
+		}
+
 		var resultField string
 
 		switch valField.Interface().(type) {
--- a/pagination.go
+++ b/pagination.go
@@ -9,6 +9,7 @@ import (
 	"crypto/sha256"
 	"encoding/json"
 	"fmt"
+	"reflect"
 	"strconv"
 
 	"github.com/go-resty/resty/v2"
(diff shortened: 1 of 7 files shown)


lgarber-akamai:
@wbh1 How would you feel about an implementation using structs and query tags? 
https://github.com/linode/linodego/commit/43bc40244851bc45bf9fbbf7e31ce91b0b605c93

zliang-akamai:
@lgarber-akamai Is this PR ready yet?

jriddle-linode (approved):

pagination.go
```diff
@@ -102,3 +124,60 @@ func (c *Client) listHelper(ctx context.Context, pager PagedResponse, opts *List
...
+
+// flattenQueryStruct flattens a structure into a Resty-compatible query param map.
+// Fields are mapped using the `query` struct tag.
+func flattenQueryStruct(val any) (map[string]string, error) {
+	result := make(map[string]string)
+
+	reflectVal := reflect.ValueOf(val)
+
+	// Deref pointer if necessary
+	if reflectVal.Kind() == reflect.Ptr {
```
zliang-akamai, line 136:
```suggestion
	if reflectVal.Kind() == reflect.Pointer {
```
[Ptr is the old name for the Pointer kind. ](https://pkg.go.dev/reflect#pkg-constants)

pagination.go
```diff
@@ -102,3 +124,60 @@ func (c *Client) listHelper(ctx context.Context, pager PagedResponse, opts *List
...
+		reflectVal = reflect.Indirect(reflectVal)
```
zliang-akamai, line 137: Can we handle `nil` pointer here?

pagination.go
```diff
@@ -102,3 +124,60 @@ func (c *Client) listHelper(ctx context.Context, pager PagedResponse, opts *List
...
+	}
+
+	valType := reflectVal.Type()
```
zliang-akamai, line 140:
```suggestion

	if reflectVal.Kind() != reflect.Struct {
		return result, fmt.Errorf(
			"expected struct type for the QueryParams but got: %s",
			reflectVal.Kind().String(),
		)
	}

	valType := reflectVal.Type()
```

pagination.go
```diff
@@ -102,3 +124,60 @@ func (c *Client) listHelper(ctx context.Context, pager PagedResponse, opts *List
...
+			return nil, fmt.Errorf("invalid query param tag: %s", currentField.Name)
+		}
+
+		// Skip if it's a zero value
+		if valField.IsZero() {
+			continue
+		}
+
+		// Deref the pointer is necessary
+		if valField.Kind() == reflect.Ptr {
```
zliang-akamai, line 162:
```suggestion
		if valField.Kind() == reflect.Pointer {
```

pagination.go
```diff
@@ -102,3 +124,60 @@ func (c *Client) listHelper(ctx context.Context, pager PagedResponse, opts *List
...
+		switch valField.Interface().(type) {
+		case string:
+			resultField = valField.String()
+		case int64, int:
+			resultField = strconv.FormatInt(valField.Int(), 10)
+		case bool:
+			resultField = strconv.FormatBool(valField.Bool())
+		default:
+			return nil, fmt.Errorf("unsupported query param type: %s", valField.Type().Name())
+		}
```
zliang-akamai, line 177: Do we want support slice (array) in query parameters? Maybe not today?

lgarber-akamai: I don't think it's necessary in this case since query params generally aren't expected to be provided in list format 👍

lgarber-akamai:
@zliang-akamai Thanks for the suggestions!

zliang-akamai:
@lgarber-akamai Are we good to merge this?

pagination.go
zliang-akamai: We may want to switch to reflect types, like we did in TF filter.

zliang-akamai (changes requested):
I can make this change later today

zliang-akamai (approved):

lgarber-akamai (approved):

zliang-akamai (approved):

PR #329, opened 2023-05-17
Dropped `LongviewSubscription` from `AccountSettingsUpdateOptions`

ezilber-akamai:
## 📝 Description

Dropped `LongviewSubscription` from `AccountSettingsUpdateOptions`.

### DIFF
--- a/account_settings.go
+++ b/account_settings.go
@@ -28,6 +28,10 @@ type AccountSettingsUpdateOptions struct {
 	// The default backups enrollment status for all new Linodes for all users on the account.  When enabled, backups are mandatory per instance.
 	BackupsEnabled *bool `json:"backups_enabled,omitempty"`
 
+	// A plan name like "longview-3"..."longview-100", or a nil value for to cancel any existing subscription plan.
+	// DEPRECATED: Use PUT /longview/plan instead to update the LongviewSubscription
+	LongviewSubscription *string `json:"longview_subscription,omitempty"`
+
 	// The default network helper setting for all new Linodes and Linode Configs for all users on the account.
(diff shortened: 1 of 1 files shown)


yec-akamai:
Why do we need to remove `LongviewSubscription` from the update options? It seems `longview_subscription` is still available in the api doc https://www.linode.com/docs/api/account/#account-settings-update

ezilber-akamai:
> Why do we need to remove `LongviewSubscription` from the update options? It seems `longview_subscription` is still available in the api doc https://www.linode.com/docs/api/account/#account-settings-update

According to the [ticket description](https://jira.linode.com/browse/TPT-2015) `LongviewSubscription` can no longer be updated through this endpoint.

yec-akamai (approved):

lgarber-akamai:
Just to add a bit of context, this field can no longer be sent to the API since the API now returns the following error if `longview_subscription` is specified: 
```
longview_subscription editing at this location is no longer available. Please use the /longview/plan endpoint instead.
```

It might still be worth keeping the field to prevent introducing a breaking change for downstream consumers, but it might be worth considering to notify users of the non-functional endpoint. Does anyone have any thoughts on this? 🙂

yec-akamai:
Maybe we can highlight this change in the corresponding release and let the user know the potential breaking change or the API doesn't work this way anymore. I'm also thinking we should let the API doc team know and update the doc as well to reduce confusion.

ezilber-akamai:
I agree that we should probably keep the field to avoid any breaking changes, and I think adding a note to the release is a great idea!

PR #338, opened 2023-06-20
fix: password_created type in account_users

yec-akamai:
## 📝 Description

A small fix to the `password_created` type. Use the `Time` type instead.

### DIFF
--- a/test/integration/account_users_test.go
+++ b/test/integration/account_users_test.go
@@ -4,11 +4,14 @@ import (
 	"context"
 	"testing"
 
+	"github.com/google/go-cmp/cmp/cmpopts"
 	"github.com/linode/linodego"
 	. "github.com/linode/linodego"
 )
 
 const usernamePrefix = "linodegotest-"
+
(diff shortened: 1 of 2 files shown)


lgarber-akamai (approved):

zliang-akamai (approved):

test/integration/account_users_test.go
```diff
@@ -4,12 +4,15 @@ import (
...
 	"testing"
 
+	"github.com/google/go-cmp/cmp/cmpopts"
 	"github.com/linode/linodego"
 	. "github.com/linode/linodego"
 )
 
 const usernamePrefix = "linodegotest-"
 
+var ignoreUserTimestampes = cmpopts.IgnoreFields(linodego.User{}, "PasswordCreated")
```
zliang-akamai, line 14: What does this var do?

yec-akamai: I meant to flag that we ignore comparing `password_created`, because it can be nullable. I found similar usage in our code to ignore comparing the timestamp.

PR #347, opened 2023-07-17
Adding missing GPU field from InstanceType and Instance

amisiorek-akamai:
## 📝 Description

Added missing GPU field to InstanceType and Instance.

## ✔️ How to Test

Added another line to existing test to check if GPU field was being successfully added to Instance object.

### DIFF
--- a/instances.go
+++ b/instances.go
@@ -64,6 +64,7 @@ type InstanceSpec struct {
 	Memory   int `json:"memory"`
 	VCPUs    int `json:"vcpus"`
 	Transfer int `json:"transfer"`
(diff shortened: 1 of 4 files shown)


yec-akamai:
Nice work so far! 

Can you run both unit and integration test to test the change? You probably need to update the fixture to pass the integration test: `make ARGS="-run TestInstances_List" fixtures`

jriddle-linode (approved):

yec-akamai (approved):

zliang-akamai:
Let's run an integration test on GitHub

PR #349, opened 2023-07-24
new: Add support for polling on events with a secondary entity ID

lgarber-akamai:
## 📝 Description

This change adds support for polling on events while filtering on both a primary and secondary entity ID. This includes a new `NewEventPollerWithSecondary(...)` function and a new `SecondaryEntityID` event poller field.

NOTE: The filtering logic for secondary entity IDs has been implemented on the client as this functionality is not currently offered by the API.

## ✔️ How to Test

```
make ARGS="-run TestEventPoller_Secondary" fixtures
```


zliang-akamai (approved):

waitfor.go
```diff
@@ -645,6 +665,10 @@ func (p *EventPoller) WaitForLatestUnknownEvent(ctx context.Context) (*Event, er
 			}
 
 			for _, event := range events {
+				if !eventMatchesSecondary(p.SecondaryEntityID, event) {
+					continue
+				}
+
```
zliang-akamai, line 671: Is adding an additional API filter when `SecondaryEntityID != nil` a better and workable idea?

PR #356, opened 2023-08-11
new: Add `betas` to support self serve beta program

yec-akamai:
## 📝 Description

Beta program is a new product or service that's not generally available to all customers. User with permissions can enroll into a beta program and access the functionalities.

Add `betas` to support listing active beta programs and getting details of a single beta program. 
Two endpoints are implemented in this PR: 
```
GET /beta
GET /betas/:id
```

JIRA: https://jira.linode.com/browse/TPT-2181 

Disable `musttag` as a workaround to avoid an issue that metalinter is getting malformed json: https://github.com/linode/linodego/actions/runs/5837079890/job/15831940532 

## ✔️ How to Test
```
make testunit
make testint
```


betas.go
```diff
@@ -0,0 +1,99 @@
...
+}
+
+// UnmarshalJSON implements the json.Unmarshaler interface
+func (beta *BetaProgram) UnmarshalJSON(b []byte) error {
+	type Mask BetaProgram
+
+	p := struct {
+		*Mask
+		StartDT *parseabletime.ParseableTime `json:"start_dt"`
+		EndDT   *parseabletime.ParseableTime `json:"end_dt"`
```
lgarber-akamai, line 53:
```suggestion
		Started *parseabletime.ParseableTime `json:"started"`
		Ended   *parseabletime.ParseableTime `json:"ended"`
```

lgarber-akamai (changes requested):
Just one small change and this should be good to go!

betas.go
yec-akamai: Makes sense. I used StartDT and EndDT is because the sample response in API spec doc use them. I'll double check with the response from alpha.

lgarber-akamai (approved):
Looks good, nice work!

zliang-akamai:
The change looks good! But do we want to merge this into `main` before it's available in production?

yec-akamai:
> The change looks good! But do we want to merge this into `main` before it's available in production?

Good question! The self serve beta endpoints are actually available in production. We test against alpha is because there isn't data (active beta programs) in production for now. It shouldn't block us from merging the change. On the other hand, we are not always going to have available betas anyway. So I think it's good to merge the implementation now.

zliang-akamai (approved):

PR #359, opened 2023-08-16
Not to override `LINODE_URL` in `Makefile`

zliang-akamai:
This is useful for testing against Alpha. Use the environmental variable `LINODE_URL` when available instead of always override it.


lgarber-akamai (approved):
Nice catch!

jriddle-linode (approved):

PR #364, opened 2023-08-22
Add `firewall_id` field for NodeBalancers

amisiorek-akamai:
## 📝 Description

Adds firewall_id to NodeBalancer struct

## ✔️ How to Test

Uncomment changes related to firewall ids in setupNodeBalancer. Currently cannot have test with firewall as there is no availability in the region with api.dev and a null point error when we use api.linode.

`make ARGS="-run TestNodeBalancer_Create_create" fixtures`


zliang-akamai:
I think this field is belong to `NodeBalancerCreateOptions` struct, according to the API design. Am I correct?

test/integration/nodebalancers_test.go
```diff
@@ -85,11 +86,15 @@ func setupNodeBalancer(t *testing.T, fixturesYaml string) (*linodego.Client, *li
 	t.Helper()
 	var fixtureTeardown func()
 	client, fixtureTeardown := createTestClient(t, fixturesYaml)
+	client, firewall, firewallTeardown, err := setupFirewall(t, []firewallModifier{}, "fixtures/TestNodeBalancer_Create")
 	createOpts := linodego.NodeBalancerCreateOptions{
 		Label:              &label,
```
zliang-akamai, line 91: `setupNodeBalancer` function is shared with different test cases, so we can't use the static string here as the fixture name. And, I am not sure if calling 2 setup functions here will work since they may override the fixture files generated by each other...

zliang-akamai (changes requested):

test/integration/nodebalancers_test.go
amisiorek-akamai: Fixing this now!

nodebalancer.go
```diff
@@ -32,6 +32,9 @@ type NodeBalancer struct {
 	// An array of tags applied to this object. Tags are for organizational purposes only.
 	Tags []string `json:"tags"`
 
+	// The ID of an associated firewall with this NodeBalancer.
+	FirewallID int `json:"firewall_id"`
+
```
zliang-akamai, line 37: I think if the returning API object doesn't include the firewall ID, we shouldn't put it in the NodeBalancer struct because some functions like `GetNodeBalancer` wouldn't know the firewall ID and will leave this field empty (value `0`). I am not 100% confident about this though, and I wonder how we handle a similar situation for other existing structs.

amisiorek-akamai: We can wait on confirmation from the API team maybe? These changes can't be merged into main either way so I don't think it would hurt to follow-up.

zliang-akamai: Yea, but they confirmed we won't have that in the response...

test/integration/nodebalancers_test.go
```diff
@@ -85,11 +85,15 @@ func setupNodeBalancer(t *testing.T, fixturesYaml string) (*linodego.Client, *li
 	t.Helper()
 	var fixtureTeardown func()
 	client, fixtureTeardown := createTestClient(t, fixturesYaml)
+	firewall, firewallTeardown, err := createFirewall(t, client)
+	defer firewallTeardown()
 	createOpts := linodego.NodeBalancerCreateOptions{
 		Label:              &label,
-		Region:             getRegionsWithCaps(t, client, []string{"NodeBalancers"})[0],
+		Region:             getRegionsWithCaps(t, client, []string{"NodeBalancers"})[6],
```
zliang-akamai, line 92:
```suggestion
		Region:             getRegionsWithCaps(t, client, []string{"NodeBalancers"})[0],
```
Maybe put 0 here to avoid index out of bounds issue?

amisiorek-akamai: Ah this is what I meant in the PR comment. If we go this route of having [6], it will pass locally. But if we put [0] it won't pass locally due to regional unavailability.

test/integration/nodebalancers_test.go
```diff
@@ -85,11 +85,15 @@ func setupNodeBalancer(t *testing.T, fixturesYaml string) (*linodego.Client, *li
 	t.Helper()
 	var fixtureTeardown func()
 	client, fixtureTeardown := createTestClient(t, fixturesYaml)
+	firewall, firewallTeardown, err := createFirewall(t, client)
+	defer firewallTeardown()
```
zliang-akamai, line 89: The deferred function call will execute right after this function, and thus we probably don't want to do it in a setup function. Imagine that a test case call this setup function, and then the the teardown function will be executed right after the setup function, and then the test case won't be able to get the resource for testing. The recommended way is to put the `firewallTeardown` function call in the teardown function defined below (line 102-107).

You can also refer to this pattern in firewall test file. https://github.com/linode/linodego/blob/main/test/integration/firewalls_test.go#L161-L164

test/integration/nodebalancers_test.go
zliang-akamai: `[0]` will be fine with the production API, and you can modify `getRegionsWithCaps` to make it only return the available region in the dev environment.

jriddle-linode (approved):
Looks good after you implement @zliang-akamai feedback.

zliang-akamai (approved):
Nice work! 🎉

PR #365, opened 2023-08-22
fix: add missing events

jriddle-linode:
## 📝 Description

Adding missing events from https://www.linode.com/docs/api/account/#events-list


account_events.go
```diff
@@ -55,101 +55,132 @@ type EventAction string
...
+	ActionAccountUpdate                 EventAction = "account_update"
+	ActionAccountSettingsUpdate         EventAction = "account_settings_update"
+	ActionBackupsEnable                 EventAction = "backups_enable"
+	ActionBackupsCancel                 EventAction = "backups_cancel"
+	ActionBackupsRestore                EventAction = "backups_restore"
+	ActionCommunityQuestionReply        EventAction = "community_question_reply"
+	ActionCommunityLike                 EventAction = "community_like"
+	ActionCreateCardUpdated             EventAction = "credit_card_updated"
+	ActionDatabaseCreate                EventAction = "database_create"
+	ActionDatabaseDegraded              EventAction = "database_degraded"
```
amisiorek-akamai, line 67: While we're updating Actions, it looks like these database_method actions aren't listed in the API for Event View?

zliang-akamai:
Can confirm the gosec scan alert is a false positive.

account_events.go
zliang-akamai: Yes, they are missing in the doc, but they exist in this repo before this PR, lol. Let's ask the doc team to add them.

yec-akamai (approved):

PR #367, opened 2023-08-24
new: Implement region-specific pricing

lgarber-akamai:
## 📝 Description

This PR implements the necessary client changes for region-specific pricing.

NOTE: These tests for this PR are _not_ expected to change until this feature is available through the API.

## ✔️ How to Test

```
make ARGS="-run TestTypes_RegionSpecific" testacc
```


jriddle-linode (approved):

types.go
```diff
@@ -10,18 +10,19 @@ import (
...
-	NetworkOut int             `json:"network_out"`
-	Memory     int             `json:"memory"`
-	Transfer   int             `json:"transfer"`
-	VCPUs      int             `json:"vcpus"`
-	GPUs       int             `json:"gpus"`
-	Successor  string          `json:"successor"`
+	ID           string              `json:"id"`
+	Disk         int                 `json:"disk"`
+	Class        LinodeTypeClass     `json:"class"` // enum: nanode, standard, highmem, dedicated, gpu
+	Price        *LinodePrice        `json:"price"`
```
yec-akamai, line 16: Just curious, since we have region-specific pricing, is `Price` going to be deprecated?

yec-akamai (approved):
Looks good!

types.go
lgarber-akamai: Good question! As of now I think the `region_prices` field will only contain exceptions and the price field will dictate the price for all other regions.

PR #373, opened 2023-08-30
Fix instance create opts

zliang-akamai:
Instance creation options should now use `InstanceConfigInterfaceCreateOptions` for the interfaces.

Also fixed a test issue.

Special thanks to @yec-akamai for discovering this issue!

### DIFF
--- a/instances.go
+++ b/instances.go
@@ -107,23 +107,23 @@ type InstanceMetadataOptions struct {
 
 // InstanceCreateOptions require only Region and Type
 type InstanceCreateOptions struct {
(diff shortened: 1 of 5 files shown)


jriddle-linode (approved):

lgarber-akamai (approved):

yec-akamai (approved):
Looks good! Is the test failure expected?

zliang-akamai:
@yec-akamai Yep, the test fixtures haven't been added due to the fact that beta is not available yet.

PR #381, opened 2023-09-13
Project: Virtual Private Cloud

zliang-akamai:
This PR is for adding VPC related functionalities to `linodego`.

All related test cases can be ran by:
`make fixtures ARGS="-run TestVPC"`
`make fixtures ARGS="-run TestInstance_ConfigInterface"`

Thanks everyone for the effort to support VPC in `linodego`!


test/integration/instance_config_test.go
```diff
@@ -0,0 +1,431 @@
...
+	config, err = client.UpdateInstanceConfig(context.Background(), instance.ID, config.ID, updateConfigOpts)
+	if err != nil {
+		teardown()
+		t.Fatal(err)
+	}
+
+	return client, vpc, vpcSubnet, instance, config, teardown
+}
+
+func TestInstance_ConfigInterfaces_AppendDelete(t *testing.T) {
```
lgarber-akamai, line 104: After re-running the fixtures, I noticed this test was failing when running `make SKIP_LINT=1 test` with a missing interaction error. As far as I can tell, this is because [fixturesTeardown](https://github.com/linode/linodego/blob/77af41ef44271e8302df4829764c9eb26d9450cf/test/integration/instances_test.go#L537) is called before the [VPC teardown] is called, which is causing the VPC deletion to not be saved in the fixtures file.

lgarber-akamai: As a not-so-pretty workaround, I was able to fix this issue by swapping the order of these two lines: https://github.com/linode/linodego/blob/b38fbd67135b16e6cd75ac9e59c3159d5daee5cc/test/integration/instance_config_test.go#L49-L50

zliang-akamai: Thank you! I think we have resolved this.

lgarber-akamai (approved):
LGTM, awesome work everyone 🎉

jriddle-linode (approved):

PR #394, opened 2023-10-05
refactor to NewError to not log fatal or panic, add tests

### LINKED ISSUE #393: [Feature]: Don't panic/fatal in NewError
### Description

This code is likely rarely reached: https://github.com/linode/linodego/blob/main/errors.go#L131-L132 but the log fatal will kill your program and someone using the package will not likely expect that behavior. e.g. `linodego.NewError(struct{}{})` will full stop exit the code

I suggest if the type not able to be handled we return that as an error and let the callee handle it as would be expected.

### Example Code

_No response_

luthermonson:
## 📝 Description

closes #393 

## ✔️ How to Test

## 📷 Preview

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -2,20 +2,21 @@ package linodego
 
 import (
 	"fmt"
-	"log"
 	"net/http"
+	"reflect"
 	"strings"
 
 	"github.com/go-resty/resty/v2"
 )
 
 const (
+	ErrorUnsupported = iota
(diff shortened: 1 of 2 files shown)


lgarber-akamai (approved):
Looks good! Could you remove the redundant `fmt.Sprintf` to get the linter passing?

amisiorek-akamai (approved):

lgarber-akamai:
@luthermonson Thanks for the contribution!

PR #396, opened 2023-10-05
fix: upload test results only on main push

ykim-1:
## 📝 Description

Quick fix to upload test results only on main push and not on PRs

## ✔️ How to Test

## 📷 Preview


lgarber-akamai (approved):

lgarber-akamai (approved):

zliang-akamai (approved):
This looks good to me!
I think in the future we can make another `job` and pass the JSON value to there from this `job` and add the if statement there to filter out non-main-push run. I think it can be prettier that way.

PR #400, opened 2023-10-07
updating errors coverage to 100%

luthermonson:
## 📝 Description
There was a type error in my last test with `*error` vs. `error` which I missed so I fixed that and added tests for the rest of the file. Note to reviewer... 
* I changed the logic in `coupleAPIErrors` to return earlier instead of putting all the coupling logic inside an if block, spiritually the same but easier to follow.
* I commented the `client.SetDebug(true)` in generic html error because it wasn't really benefiting anything and making the test output spammy. if you want it back, I'll happily uncomment it

Rewrites the errors tests to give 100% coverage, because... tests.

## ✔️ How to Test

`make testunit`

## 📷 Preview

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -47,41 +47,45 @@ type APIError struct {
 
 func coupleAPIErrors(r *resty.Response, err error) (*resty.Response, error) {
 	if err != nil {
+		// an error was raised in go code, no need to check the resty Response
 		return nil, NewError(err)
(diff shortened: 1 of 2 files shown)


jriddle-linode (approved):
Looks good!

zliang-akamai (approved):

PR #408, opened 2023-10-16
Fix `ipv4` Field of Linode Instance Interface

zliang-akamai:
## 📝 Description

Omit any empty field during JSON marshaling in IPv4 struct. 

## ✔️ How to Test

Run `make ARGS="-run TestInstance_Config" fixtures` against alpha environment


lgarber-akamai (approved):
Looks good and tests pass!

jriddle-linode (approved):

PR #426, opened 2023-11-15
VPC Test Create Helper Function Update

zliang-akamai:
## 📝 Description

- Make `createVPCWithSubnet` function to create VPC subnet via the VPC creation API call, as a replacement to the 2 API calls style of creating a subnet (one call creates VPC and another creates the subnet).
- Add `createSubnetInVPC` function that create a subnet in a given VPC.

## ✔️ How to Test
`make fixtures ARGS="-run TestVPC"`
`make fixtures ARGS="-run TestInstance_Config"`


amisiorek-akamai (approved):
Tests work locally!

jriddle-linode (approved):

PR #428, opened 2023-11-17
Fields for Unified Migration Project

zliang-akamai:
## 📝 Description

This is to support the feature from the unified migration project. Tests are failing because lack of fixture, which we can generate when the change is in the production.

## ✔️ How to Test

`make ARGS="-run TestInstance_Resize" fixtures` against alpha.
Don't forget making get regions function only to return `us-east`.

We can't test migrate in the test cases because it requires an admin portal operation prior to the API call, and it don't return any useful value, which makes mocked response ruled out. And it's just a field, should be very easy to test it manually.

### DIFF
--- a/instances.go
+++ b/instances.go
@@ -35,6 +35,13 @@ const (
 	InstanceResizing     InstanceStatus = "resizing"
 )
 
+type InstanceMigrationType string
+
+const (
+	WarmMigration InstanceMigrationType = "warm"
+	ColdMigration InstanceMigrationType = "cold"
+)
+
 // Instance represents a linode object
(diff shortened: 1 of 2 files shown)


lgarber-akamai (approved):
Running fixtures passes as expected after overriding `LINODE_URL` and explicitly using `us-east` as the testing region. Nice work!

jriddle-linode (approved):
LGTM. thanks for the tip Lena!

zliang-akamai:
Thanks for the review! Note that test fixtures will be added once we get this feature on the production server.

PR #430, opened 2023-11-20
test: add checks for test result fields in upload script for TOD

ykim-1:
## 📝 Description
There is an issue on TOD side when taking xml reports generated from our Go repositories.

Cause of the issue is that TOD is only expecting 1 testsuite variable under testsuites but the current xmls have multiple instances of testsuite.

This PR is to add some checks for fields related to test result before processing and uploading xml to TOD

## ✔️ How to Test

Need to monitor TOD side if results got uploaded successfully

## 📷 Preview

### DIFF
--- a/scripts/test_report_upload_script.py
+++ b/scripts/test_report_upload_script.py
@@ -23,10 +23,10 @@ def change_xml_report_to_tod_acceptable_version(file_name):
     testsuites_element = root
 
     # total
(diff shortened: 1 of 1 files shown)


zliang-akamai (approved):

lgarber-akamai (approved):
LGTM!

PR #432, opened 2023-12-01
feat(errors): added error.Is compatibility

shanduur:
## 📝 Description

This pull request introduces the implementation of the [internal interface from `errors.Is`](https://cs.opensource.google/go/go/+/refs/tags/go1.21.4:src/errors/wrap.go;l=54-56) for the custom `linodego.Error` struct. The `Is` method is implemented to compare the relevant fields (in this case only the status code) of the struct for equality. This enhancement aims to improve the usability of the `errors.Is` function when working with instances of the `linodego.Error` type.

## ✔️ How to Test

I added table-driven tests for testing `errors.Is` compatibility. I tried to find edge cases, thus table is quite big.

Runing only newly added tests:

```sh
go test -race -run ^TestErrorIs$ github.com/linode/linodego
```


errors.go
```diff
@@ -11,11 +12,11 @@ import (
...
 const (
 	ErrorUnsupported = iota
-	// ErrorFromString is the Code identifying Errors created by string types
+	// ErrorFromString is the Code identifyinerr Errors created by string types
 	ErrorFromString
-	// ErrorFromError is the Code identifying Errors created by error types
+	// ErrorFromError is the Code identifyinerr Errors created by error types
 	ErrorFromError
-	// ErrorFromStringer is the Code identifying Errors created by fmt.Stringer types
+	// ErrorFromStringer is the Code identifyinerr Errors created by fmt.Stringer types
```
zliang-akamai, line 19: Is this an intentional change?

shanduur: Not at all! Thanks for pointing this out, I reverted that change.

lgarber-akamai (approved):
LGTM, thanks for the contribution!

zliang-akamai (approved):
It looks good to me, and tests were passed. Thank you for the contribution!

PR #435, opened 2023-12-04
new: add support listing `instance` `firewalls`

jriddle-linode:
## 📝 Description

Add support for `instances/{linode_id}/firewalls` endpoint

## ✔️ How to Test

```bash
make ARGS='-run TestInstanceFirewalls_List' fixtures
```

### DIFF
--- a/test/integration/instance_firewalls_test.go
+++ b/test/integration/instance_firewalls_test.go
@@ -42,8 +42,8 @@ func setupInstanceFirewall(t *testing.T, firewallModifiers []firewallModifier, f
 	firewall, firewallTeardown, err := createFirewall(t, client, firewallModifiers...)
 
 	teardown := func() {
-		instanceTeardown()
 		firewallTeardown()
+		instanceTeardown()
 	}
 	return client, instance, firewall, teardown, err
 }
--- a/test/integration/fixtures/TestInstanceFirewalls_List.yaml
+++ b/test/integration/fixtures/TestInstanceFirewalls_List.yaml
@@ -251,9 +251,9 @@ interactions:
     url: https://api.linode.com/v4beta/linode/instances
     method: POST
   response:
-    body: '{"id": 52733363, "label": "linodego-fw-inst-test", "group": "", "status":
+    body: '{"id": 52739346, "label": "linodego-fw-inst-test", "group": "", "status":
       "provisioning", "created": "2018-01-02T03:04:05", "updated": "2018-01-02T03:04:05",
-      "type": "g6-nanode-1", "ipv4": ["172.105.41.60"], "ipv6": "1234::5678/128",
+      "type": "g6-nanode-1", "ipv4": ["172.105.49.53"], "ipv6": "1234::5678/128",
       "image": "linode/debian9", "region": "ap-west", "specs": {"disk": 25600, "memory":
       1024, "vcpus": 1, "gpus": 0, "transfer": 1000}, "alerts": {"cpu": 90, "network_in":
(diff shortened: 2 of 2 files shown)


instance_firewalls.go
```diff
@@ -0,0 +1,41 @@
...
+	if err != nil {
+		return 0, 0, err
+	}
+	castedRes := res.Result().(*InstanceFirewallsPagedResponse)
+	resp.Data = append(resp.Data, castedRes.Data...)
+	return castedRes.Pages, castedRes.Results, nil
+}
+
+// ListInstanceFirewalls returns a paginated list of Cloud Firewalls for linodeID
+func (c *Client) ListInstanceFirewalls(ctx context.Context, opts *ListOptions, linodeID int) ([]Firewall, error) {
```
lgarber-akamai, line 32: Same as https://github.com/linode/linodego/pull/436#discussion_r1417591956

test/integration/instance_firewalls_test.go
```diff
@@ -0,0 +1,49 @@
...
+	device := linodego.DevicesCreationOptions{Linodes: []int{instance.ID}}
+	firewallModifiers = append(firewallModifiers,
+		func(createOpts *linodego.FirewallCreateOptions) {
+			createOpts.Devices = device
+		})
+	firewall, firewallTeardown, err := createFirewall(t, client, firewallModifiers...)
+
+	teardown := func() {
+		instanceTeardown()
+		firewallTeardown()
```
lgarber-akamai, line 46: Due to some fixtures weirdness `setupInstance`'s teardown will [close the fixtures file](https://github.com/linode/linodego/blob/main/test/integration/integration_suite_test.go#L154C18-L154C18), so you'll need to run it after `firewallTeardown()` so the firewall DELETE request is stored in the fixture file. Alternatively, you could directly call [createTestClient](https://github.com/linode/linodego/blob/main/test/integration/integration_suite_test.go#L120) here and use [createInstance](https://github.com/linode/linodego/blob/main/test/integration/instances_test.go#L446C6-L446C20) rather than `setupInstance`.

This should fix the CI failures after re-running the fixtures :+1:

lgarber-akamai (changes requested):
Just one test change and this should be good go to :slightly_smiling_face:

test/integration/instance_firewalls_test.go
jriddle-linode: Thank you for the explanation.

lgarber-akamai (approved):
LGTM, great work! 

Tested manually using the following:

```go
const testLabel = "firewall-test"

func main() {

	inst, err := client.CreateInstance(
		context.Background(),
		linodego.InstanceCreateOptions{
			Label:  testLabel,
			Region: "us-mia",
			Type:   "g6-nanode-1",
		},
	)
	if err != nil {
		log.Fatal(err)
	}

	fw, err := client.CreateFirewall(
		context.Background(),
		linodego.FirewallCreateOptions{
			Label: testLabel,
			Rules: linodego.FirewallRuleSet{
				InboundPolicy:  "DROP",
				OutboundPolicy: "DROP",
			},
			Devices: linodego.DevicesCreationOptions{
				Linodes: []int{inst.ID},
			},
		},
	)
	if err != nil {
		log.Fatal(err)
	}

	instFWs, err := client.ListInstanceFirewalls(
		context.Background(),
		inst.ID,
		nil,
	)
	if err != nil {
		log.Fatal(err)
	}

	fmt.Println(instFWs[0].ID, fw.ID)
}
```

yec-akamai (approved):
Tests passed locally. Nice work!

PR #450, opened 2024-01-16
Pass VPCID for instance creation

mhmxs:
## 📝 Description

Linode API supports adding an instance to an existing VPC, but without this change it is not possible via linodego.

## ✔️ How to Test

Create a VPC
Create an instance inside the VPC

Unit test has been fixed.

## 📷 Preview


zliang-akamai:
I think we can use the `SubnetID` field to associate the interface with the VPC. https://github.com/linode/linodego/pull/450/files#diff-f1f86fccc61d71ddf6668c2fae324f041e279c2df9f736e032ed25a85a442253L33

mhmxs:
> I think we can use the `SubnetID` field to associate the newly created interface with a VPC subnet. https://github.com/linode/linodego/pull/450/files#diff-f1f86fccc61d71ddf6668c2fae324f041e279c2df9f736e032ed25a85a442253L33

@zliang-akamai If we check the web UI, where the functionality is available it sends both:

image

Do you mean to find vpc id by subnet id?

PR #451, opened 2024-01-17
Make multiple fields nullable (nilable) in networking interface

zliang-akamai:
## 📝 Description

The API schema change made some fields nullable now. `linodego` change in this PR will work with both the alpha and the production, but some tests will be failed before the API schema change deployed to the production. We may want to hold on for now, but feel free to review though. Fixtures will be added after the production deployment.

## ✔️ How to Test

`make ARGS="-run TestInstance_Config" fixtures`

### DIFF
--- a/instance_config_interfaces.go
+++ b/instance_config_interfaces.go
@@ -16,13 +16,13 @@ type InstanceConfigInterface struct {
 	Active      bool                   `json:"active"`
 	VPCID       *int                   `json:"vpc_id"`
 	SubnetID    *int                   `json:"subnet_id"`
-	IPv4        VPCIPv4                `json:"ipv4"`
+	IPv4        *VPCIPv4               `json:"ipv4"`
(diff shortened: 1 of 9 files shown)


yec-akamai (commented):
I ran the intergration tests locally but keep getting a failed case:
```
    instance_config_test.go:311: failed to update linode interfaces: configs do not match
    instance_config_test.go:328: failed to update linode interfaces: configs do not match
--- FAIL: TestInstance_ConfigInterfaces_Update (0.00s)
```
Do you mind double checking the case?

zliang-akamai:
Hi @yec-akamai, thanks for the review! Did you test it with production API? We will have to wait until next week to get these passed when the API changed deployed to production. You can try testing it with alpha and these errors should be gone.

yec-akamai (approved):
Tests passed successfully! Nice work!

ykim-1 (approved):
Nice and tests all passed locally

PR #458, opened 2024-02-28
fix: Exclude request body in POST and PUT helpers when options are unspecified

lgarber-akamai:
## 📝 Description

This change fixes a small issue that made the POST and PUT request helpers encode empty bodies as `null`, which is rejected by the API. This works by making the options argument variadic and only calling `r.SetBody(...)` if the argument is specified.

## ✔️ How to Test

### Unit Testing

```
make testunit
```

### Integration Testing

NOTE: This test calls out to the `client.ShutdownInstance(...)` function, which consumes this helper and does not accept a request body.

```
make ARGS="-run TestEventPoller_InstancePower" fixtures
```

### Manual Testing

```go
package main

import (
	"context"
	"log"
	"os"

	"github.com/linode/linodego"
)

func main() {
	ctx := context.Background()

	inst, err := client.CreateInstance(ctx, linodego.InstanceCreateOptions{
		Region:   "us-mia",
		Type:     "g6-nanode-1",
		Label:    "test-instance",
		Image:    "linode/ubuntu22.04",
		RootPass: "v3rys3cur3t3stp4ssw0rd!!!!",
	})
	if err != nil {
		log.Fatal(err)
	}

	inst, err = client.WaitForInstanceStatus(ctx, inst.ID, linodego.InstanceRunning, 180)
	if err != nil {
		log.Fatal(err)
	}

	// Attempt to shutdown the instance
	// This calls the request helper with no options internally
	if err := client.ShutdownInstance(ctx, inst.ID); err != nil {
		log.Fatal(err)
	}
}
```
2. Ensure no errors are raised.

### DIFF
--- a/instances.go
+++ b/instances.go
@@ -5,7 +5,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"net"
-	"net/url"
 	"time"
 
 	"github.com/go-resty/resty/v2"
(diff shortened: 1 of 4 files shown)


jriddle-linode (approved):
LGTM. passes tests locally

ykim-1 (approved):
LGTM

PR #464, opened 2024-03-04
fix: [BREAKING] Treat ip_ranges as a pointer in config interface update options

### LINKED ISSUE #462: [Bug]: unable to set ip_ranges to empty of VPC interface once set
### go Version

go version go1.21.7 linux/amd64

### API Wrapper Version

github.com/linode/linodego v1.29.0

### Code Snippet

```go
package main

import (
	"context"
	"fmt"

	"github.com/linode/linodego"
	"golang.org/x/oauth2"

	"log"
	"net/http"
	"os"
)

func main() {
	if !ok {
	}

	oauth2Client := &http.Client{
		Transport: &oauth2.Transport{
			Source: tokenSource,
		},
	}

	linodeClient.SetDebug(true)

	// Get instance configs
	instance, _ := linodeClient.GetInstance(context.Background(), 55519453)
	configs, err := linodeClient.ListInstanceConfigs(context.Background(), instance.ID, &linodego.ListOptions{})
	if err != nil {
		log.Fatal(err)
	}
	fmt.Printf("%v", configs[0].Interfaces)

	// Try clearing ip_ranges
	interfaceUpdateOptions := linodego.InstanceConfigInterfaceUpdateOptions{
		IPRanges: []string{},
	}

	// Set empty ip_ranges for VPC interface
	for _, iface := range configs[0].Interfaces {
		if iface.VPCID != nil && iface.IPv4.VPC != "" {
			resp, err := linodeClient.UpdateInstanceConfigInterface(context.Background(), instance.ID, configs[0].ID, iface.ID, interfaceUpdateOptions)
			if err != nil {
				log.Fatal(err)
			}
			fmt.Printf("Deleted routes. New output: %v", resp)
			if len(resp.IPRanges) > 0 {
				log.Printf("\nFailed to empty ip_ranges")
			}
		}
	}
}
```

### Expected Behavior

ip_ranges should be set to empty:
```
{
   "ip_ranges": []
}
```

### Actual Behavior

ip_ranges of interface connected to VPC never gets cleared.

Output sample:

Initial instance state:
```
==============================================================================
2024/03/02 05:23:05.324915 DEBUG RESTY 
==============================================================================
~~~ REQUEST ~~~
GET  /v4/linode/instances/55519453/configs  HTTP/1.1
HOST   : api.linode.com
HEADERS:
        Accept: application/json
        Content-Type: application/json
        User-Agent: linodego/v1.29.0 https://github.com/linode/linodego
BODY   :
***** NO CONTENT *****
------------------------------------------------------------------------------
~~~ RESPONSE ~~~
STATUS       : 200 OK
PROTO        : HTTP/1.1
RECEIVED AT  : 2024-03-02T05:23:05.324753224Z
TIME DURATION: 126.797914ms
BODY         :
{
   "data": [
      {
         "id": 58641324,
         "label": "My Ubuntu 22.04 LTS Disk Profile",
         "helpers": {
            "updatedb_disabled": true,
            "distro": true,
            "modules_dep": true,
            "network": true,
            "devtmpfs_automount": true
         },
         "kernel": "linode/grub2",
         "comments": "",
         "memory_limit": 0,
         "created": "2024-03-01T19:14:27",
         "updated": "2024-03-01T19:14:27",
         "root_device": "/dev/sda",
         "devices": {
            "sda": {
               "disk_id": 109845040,
               "volume_id": null
            },
            "sdb": {
               "disk_id": 109845041,
               "volume_id": null
            },
            "sdc": null,
            "sdd": null,
            "sde": null,
            "sdf": null,
            "sdg": null,
            "sdh": null
         },
         "initrd": null,
         "run_level": "default",
         "virt_mode": "paravirt",
         "interfaces": [
            {
               "id": 1137088,
               "purpose": "public",
               "primary": true,
               "active": true,
               "ipam_address": null,
               "label": null,
               "vpc_id": null,
               "subnet_id": null,
               "ipv4": null,
               "ipv6": null,
               "ip_ranges": null
            },
            {
               "id": 1137089,
               "purpose": "vpc",
               "primary": false,
               "active": true,
               "ipam_address": null,
               "label": null,
               "vpc_id": 26186,
               "subnet_id": 26443,
               "ipv4": {
                  "vpc": "10.0.0.5",
                  "nat_1_1": null
               },
               "ipv6": null,
               "ip_ranges": [
                  "10.192.3.0/24"
               ]
            }
         ]
      }
   ],
   "page": 1,
   "pages": 1,
   "results": 1
}
```

Request sent to set ip_ranges to empty:
```
==============================================================================
~~~ REQUEST ~~~
PUT  /v4/linode/instances/55519453/configs/58641324/interfaces/1137089  HTTP/1.1
HOST   : api.linode.com
HEADERS:
        Accept: application/json
        Content-Type: application/json
        User-Agent: linodego/v1.29.0 https://github.com/linode/linodego
BODY   :
{}
------------------------------------------------------------------------------
~~~ RESPONSE ~~~
STATUS       : 200 OK
PROTO        : HTTP/1.1
RECEIVED AT  : 2024-03-02T05:23:05.495718273Z
TIME DURATION: 170.160593ms
BODY         :
{
   "id": 1137089,
   "purpose": "vpc",
   "primary": false,
   "active": true,
   "ipam_address": null,
   "label": null,
   "vpc_id": 26186,
   "subnet_id": 26443,
   "ipv4": {
      "vpc": "10.0.0.5",
      "nat_1_1": null
   },
   "ipv6": null,
   "ip_ranges": [
      "10.192.3.0/24"
   ]
}
```

Ideally, in the end when we call linodeClient.UpdateInstanceConfigInterface() with empty ip_ranges, it should have set ip_ranges to empty and response should not contain ip_ranges still set.

I can delete an item from ip_ranges if there are more than 1 entries, its just that I am unable to set it to empty.

### Steps to Reproduce

1. Bring up a linode in VPC.
2. Try setting ip_ranges for the VPC interface.
3. Once set, try clearing the ip_ranges.

If one has a linode in VPC, they can use the above script to reproduce the issue. Just change the instance id with your instance id and set atleast one ip subnet in ip_ranges.

lgarber-akamai:
## 📝 Description

This change resolves an issue that prevented users from updating the interface `ip_ranges` field with an explicitly empty list. This works by treating `ip_ranges` as a pointer so it will be only omitted if the pointer is nil, as opposed to if the slice is nil OR the slice is empty. 

Resolves #462 

## ✔️ How to Test

### Integration Testing

```
make ARGS="-run TestInstance_ConfigInterface_Update" fixtures
```

### Manual Testing

```go
package main

import (
	"context"
	"fmt"
	"github.com/linode/linodego"
	"log"
	"os"
)

func main() {
	ctx := context.Background()

	vpc, err := client.CreateVPC(ctx, linodego.VPCCreateOptions{
		Label:  "test-vpc",
		Region: "us-mia",
		Subnets: []linodego.VPCSubnetCreateOptions{
			{
				Label: "test-subnet",
				IPv4:  "10.0.0.0/24",
			},
		},
	})
	if err != nil {
		log.Fatal(err)
	}

	defer func() {
		client.DeleteVPC(ctx, vpc.ID)
	}()

	inst, err := client.CreateInstance(
		ctx,
		linodego.InstanceCreateOptions{
			Region: "us-mia",
			Label:  "test-inst",
			Type:   "g6-nanode-1",
		},
	)
	if err != nil {
		log.Fatal(err)
	}
	defer func() {
		client.DeleteInstance(ctx, inst.ID)
	}()

	cfg, err := client.CreateInstanceConfig(
		ctx,
		inst.ID,
		linodego.InstanceConfigCreateOptions{
			Label:   "test-config",
			Devices: linodego.InstanceConfigDeviceMap{},
			Interfaces: []linodego.InstanceConfigInterfaceCreateOptions{
				{
					Purpose:  linodego.InterfacePurposeVPC,
					SubnetID: &vpc.Subnets[0].ID,
					IPRanges: []string{"10.0.0.5/32"},
				},
			},
		},
	)
	if err != nil {
		log.Fatal(err)
	}

	fmt.Println("ip_ranges before update:", cfg.Interfaces[0].IPRanges)

	updatedIPRanges := make([]string, 0)

	iface, err := client.UpdateInstanceConfigInterface(
		ctx,
		inst.ID,
		cfg.ID,
		cfg.Interfaces[0].ID,
		linodego.InstanceConfigInterfaceUpdateOptions{
			IPRanges: &updatedIPRanges,
		},
	)
	if err != nil {
		log.Fatal(err)
	}

	fmt.Println("ip_ranges after update: ", iface.IPRanges)
}
```

2. Ensure no errors are raised and the output looks similar to the following:

```
ip_ranges before update: [10.0.0.5/32]
ip_ranges after update:  []
```

### DIFF
--- a/instance_config_interfaces.go
+++ b/instance_config_interfaces.go
@@ -41,6 +41,27 @@ type InstanceConfigInterfaceUpdateOptions struct {
 	IPRanges []string `json:"ip_ranges,omitempty"`
 }
 
+// MarshalJSON implements the JSON marshal interface method for the interface update options.
+// This is a workaround for #462, allowing users to explicitly clear the ip_ranges field.
+func (opts InstanceConfigInterfaceUpdateOptions) MarshalJSON() ([]byte, error) {
+	type Mask InstanceConfigInterfaceUpdateOptions
+
+	// If IP ranges is nil, return the default
+	if opts.IPRanges == nil {
+		return json.Marshal(opts)
+	}
+
+	maskedOpts := &struct {
+		Mask
+		IPRanges []string `json:"ip_ranges"`
+	}{
+		Mask:     (Mask)(opts),
+		IPRanges: opts.IPRanges,
+	}
+
+	return json.Marshal(maskedOpts)
+}
+
 type InstanceConfigInterfacesReorderOptions struct {
 	IDs []int `json:"ids"`
 }
--- a/instance_config_interfaces.go
+++ b/instance_config_interfaces.go
@@ -36,30 +36,9 @@ type InstanceConfigInterfaceCreateOptions struct {
 }
 
 type InstanceConfigInterfaceUpdateOptions struct {
-	Primary  bool     `json:"primary,omitempty"`
-	IPv4     *VPCIPv4 `json:"ipv4,omitempty"`
-	IPRanges []string `json:"ip_ranges,omitempty"`
-}
-
-// MarshalJSON implements the JSON marshal interface method for the interface update options.
-// This is a workaround for #462, allowing users to explicitly clear the ip_ranges field.
-func (opts InstanceConfigInterfaceUpdateOptions) MarshalJSON() ([]byte, error) {
-	type Mask InstanceConfigInterfaceUpdateOptions
-
-	// If IP ranges is nil, return the default
-	if opts.IPRanges == nil {
-		return json.Marshal(opts)
-	}
-
-	maskedOpts := &struct {
-		Mask
-		IPRanges []string `json:"ip_ranges"`
-	}{
-		Mask:     (Mask)(opts),
-		IPRanges: opts.IPRanges,
-	}
-
-	return json.Marshal(maskedOpts)
+	Primary  bool      `json:"primary,omitempty"`
+	IPv4     *VPCIPv4  `json:"ipv4,omitempty"`
+	IPRanges *[]string `json:"ip_ranges,omitempty"`
 }
 
 type InstanceConfigInterfacesReorderOptions struct {
@@ -112,8 +91,13 @@ func (i InstanceConfigInterface) GetUpdateOptions() InstanceConfigInterfaceUpdat
 		}
 	}
 
-	if len(i.IPRanges) > 0 {
-		opts.IPRanges = i.IPRanges
+	if i.IPRanges != nil {
+		// Copy the slice to prevent accidental
+		// mutations
+		copiedIPRanges := make([]string, len(i.IPRanges))
+		copy(copiedIPRanges, i.IPRanges)
+
+		opts.IPRanges = &copiedIPRanges
 	}
 
 	return opts
--- a/test/integration/fixtures/TestInstance_ConfigInterface_Update.yaml
+++ b/test/integration/fixtures/TestInstance_ConfigInterface_Update.yaml
@@ -228,7 +228,7 @@ interactions:
       Content-Type:
       - application/json
       Expires:
-      - Wed, 14 Feb 2024 18:09:14 GMT
+      - Mon, 04 Mar 2024 20:22:16 GMT
       Pragma:
       - no-cache
       Strict-Transport-Security:
@@ -254,7 +254,7 @@ interactions:
     code: 200
     duration: ""
 - request:
-    body: '{"region":"us-iad","type":"g6-nanode-1","label":"go-test-ins-wo-disk-73be3u8uqn82","booted":false}'
+    body: '{"region":"us-iad","type":"g6-nanode-1","label":"go-test-ins-wo-disk-o27gv95tu35t","booted":false}'
     form: {}
     headers:
       Accept:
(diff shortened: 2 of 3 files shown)


zliang-akamai (commented):
Another option would be making `IPRanges` as a pointer, but that would be a breaking change for the users? I think it might be fine since we are still in the early phase of VPC? Or maybe we can schedule it to linodego v2?

lgarber-akamai:
> Another option would be making `IPRanges` as a pointer, but that would be a breaking change for the users? I think it might be fine since we are still in the early phase of VPC? Or maybe we can schedule it to linodego v2?

That's a good point! I wrote this with avoiding a breaking change in mind but it might make sense to transition this to a pointer now or in v2 for consistency with the rest of the codebase.

yec-akamai (approved):
LGTM! Tests passed locally

jriddle-linode (approved):
LGTM

PR #465, opened 2024-03-04
new: Add VPC to `InstanceIPv4Response`

zliang-akamai:
## 📝 Description

The API has been updated, and the VPC IPs of an instance is now being returned in the Networking Info List API.

## ✔️ How to Test
```bash
make ARGS="-run TestIPAddresses" fixtures
```

```go

import (
	"context"
	"fmt"
	"github.com/linode/linodego"
	"log"
	"os"
)

func main() {
	ctx := context.Background()

	client := linodego.NewClient(nil)
	client.SetToken(os.Getenv("LINODE_TOKEN"))

	vpc, err := client.CreateVPC(ctx, linodego.VPCCreateOptions{
		Label:  "test-vpc",
		Region: "us-mia",
		Subnets: []linodego.VPCSubnetCreateOptions{
			{
				Label: "test-subnet",
				IPv4:  "10.0.0.0/24",
			},
		},
	})
	if err != nil {
		log.Fatal(err)
	}

	defer func() {
		client.DeleteVPC(ctx, vpc.ID)
	}()

	inst, err := client.CreateInstance(
		ctx,
		linodego.InstanceCreateOptions{
			Region: "us-mia",
			Label:  "test-inst",
			Type:   "g6-nanode-1",
		},
	)
	if err != nil {
		log.Fatal(err)
	}
	defer func() {
		client.DeleteInstance(ctx, inst.ID)
	}()

	_, err = client.CreateInstanceConfig(
		ctx,
		inst.ID,
		linodego.InstanceConfigCreateOptions{
			Label:   "test-config",
			Devices: linodego.InstanceConfigDeviceMap{},
			Interfaces: []linodego.InstanceConfigInterfaceCreateOptions{
				{
					Purpose:  linodego.InterfacePurposeVPC,
					SubnetID: &vpc.Subnets[0].ID,
				},
			},
		},
	)
	if err != nil {
		log.Fatal(err)
	}
	i, err := client.GetInstanceIPAddresses(context.Background(), inst.ID)
	fmt.Println("VPC IP: ", *i.IPv4.VPC[0].Address)
}
```

PR #468, opened 2024-03-07
new: Implement ListVPCIPAddresses

zliang-akamai:
## 📝 Description

Need to merge [this PR](https://github.com/linode/linodego/pull/465) first then rebase this PR.

This for adding `ListVPCIPAddresses` function to the client.

## ✔️ How to Test
GHA Run: https://github.com/linode/linodego/actions/runs/8195608429
```bash
make ARGS="-run TestVPC_ListIPAddresses" fixtures
```

```go
import (
	"context"
	"encoding/json"
	"fmt"
	"os"

	"github.com/linode/linodego"
)

func main() {

	addresses, err := client.ListVPCIPAddresses(context.Background(), linodego.NewListOptions(1, "{\"linode_id\": 55774762}"))
	if err != nil {
		panic(err)
	}
	
	jsonAddresses, _ := json.Marshal(addresses)
	fmt.Println(string(jsonAddresses))
}
```


instance_ips.go
```diff
@@ -19,6 +19,7 @@ type InstanceIPv4Response struct {
 	Private  []*InstanceIP `json:"private"`
 	Shared   []*InstanceIP `json:"shared"`
 	Reserved []*InstanceIP `json:"reserved"`
+	VPC      []*VPCIP      `json:"vpc"`
```
lgarber-akamai, line 22: It's interesting that these are defined as slices of pointers 🤔 , although that's not the fault of this PR so this should be fine

lgarber-akamai (approved):
LGTM, nice work!

instance_ips.go
zliang-akamai: I am not sure.. API returns empty list `[]` instead of `null` when there is not an IP in the category. But yea, I think we can keep the pattern just to be safe.

jriddle-linode (approved):
LGTM

PR #481, opened 2024-03-20
Skip `TestBetaProgram_Get` When No Beta Program is Available

zliang-akamai:
## 📝 Description

This is to fix `TestBetaProgram_Get` test case when there is not a beta program available.

## ✔️ How to Test
```bash
make ARGS="-run TestBetaProgram_Get" fixtures
```

### DIFF
--- a/test/integration/betas_test.go
+++ b/test/integration/betas_test.go
@@ -26,8 +26,15 @@ func TestBetaPrograms_List(t *testing.T) {
 func TestBetaProgram_Get(t *testing.T) {
 	client, teardown := createTestClient(t, "fixtures/TestBetaProgram_Get")
 	defer teardown()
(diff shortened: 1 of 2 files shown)


test/integration/betas_test.go
```diff
@@ -26,8 +26,12 @@ func TestBetaPrograms_List(t *testing.T) {
 func TestBetaProgram_Get(t *testing.T) {
 	client, teardown := createTestClient(t, "fixtures/TestBetaProgram_Get")
 	defer teardown()
+	betaPrograms, err := client.ListBetaPrograms(context.Background(), linodego.NewListOptions(0, ""))
```
lgarber-akamai, line 29: I think the list options can be nil here 👍

lgarber-akamai (approved):

lgarber-akamai (approved):

PR #503, opened 2024-05-08
Linode Disk Encryption Support

avestuk:
## 📝 Description

Allows Linode instances to be created with disk encryption. 

Also handles creating LKE node pools with disk encryption enabled 

## ✔️ How to Test

Use cmd/test.go from this [branch](https://github.com/linode/linodego/pull/504) 
Add disk encryption tags to your user and `go run cmd/test.go` to create a Linode with and without encryption plus an LKE cluster with a single node pool that has encryption enabled. 

Be a little tough as there currently are none. I will ask DX for their advice here.

### DIFF
--- a/instances.go
+++ b/instances.go
@@ -43,25 +43,26 @@ const (
 
 // Instance represents a linode object
 type Instance struct {
-	ID              int             `json:"id"`
-	Created         *time.Time      `json:"-"`
-	Updated         *time.Time      `json:"-"`
-	Region          string          `json:"region"`
-	Alerts          *InstanceAlert  `json:"alerts"`
-	Backups         *InstanceBackup `json:"backups"`
-	Image           string          `json:"image"`
-	Group           string          `json:"group"`
-	IPv4            []*net.IP       `json:"ipv4"`
-	IPv6            string          `json:"ipv6"`
-	Label           string          `json:"label"`
-	Type            string          `json:"type"`
-	Status          InstanceStatus  `json:"status"`
-	HasUserData     bool            `json:"has_user_data"`
-	Hypervisor      string          `json:"hypervisor"`
-	HostUUID        string          `json:"host_uuid"`
-	Specs           *InstanceSpec   `json:"specs"`
-	WatchdogEnabled bool            `json:"watchdog_enabled"`
-	Tags            []string        `json:"tags"`
+	ID              int                    `json:"id"`
+	Created         *time.Time             `json:"-"`
+	Updated         *time.Time             `json:"-"`
+	Region          string                 `json:"region"`
+	Alerts          *InstanceAlert         `json:"alerts"`
+	Backups         *InstanceBackup        `json:"backups"`
+	Image           string                 `json:"image"`
+	Group           string                 `json:"group"`
+	IPv4            []*net.IP              `json:"ipv4"`
+	IPv6            string                 `json:"ipv6"`
+	Label           string                 `json:"label"`
+	Type            string                 `json:"type"`
+	Status          InstanceStatus         `json:"status"`
+	HasUserData     bool                   `json:"has_user_data"`
+	Hypervisor      string                 `json:"hypervisor"`
+	HostUUID        string                 `json:"host_uuid"`
+	Specs           *InstanceSpec          `json:"specs"`
+	WatchdogEnabled bool                   `json:"watchdog_enabled"`
+	Tags            []string               `json:"tags"`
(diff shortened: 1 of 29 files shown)


mrogers950 (approved):
LGTM

zliang-akamai (approved):
Works well on my end! Thanks for putting this up!

scmeggitt (approved):

instances.go
```diff
@@ -92,6 +93,17 @@ type InstanceBackup struct {
 	} `json:"schedule,omitempty"`
 }
 
+type instanceDiskEncryption string
```
lgarber-akamai, line 96: Could we make this public?

```suggestion
type InstanceDiskEncryption string
```

instances.go
```diff
@@ -92,6 +93,17 @@ type InstanceBackup struct {
...
+
+func (i *instanceDiskEncryption) String() string {
+	return string(*i)
+}
```
lgarber-akamai, line 100: Generally we don't implement `.String(...)` methods on our consts but I actually really like this pattern. It might make sense to implement it in some sort of base struct in the future :eyes:

instances.go
avestuk: We could make it public but then you can pass any string as an `InstanceDiskEncryption` type rather than just the two publically exposed vars. Those are vars rather than const as you can't reference a const.

lgarber-akamai: Gotcha, that makes sense! I don't think we have a precedent for private consts like this but it might not be a bad idea to adopt it. 

My main reservation would be if linodego fell out of parity with the API, users would be unable to specify any new consts that don't yet exist in linodego. That might be considered undefined behavior though and is probably a situation we would want to avoid in the first place :thinking:

zliang-akamai: I think it is fine to make it a public type because allowing users to create a custom string to be `InstanceDiskEncryption` won't increase any risk from SDK perspectives. And it has been a common way to handle it in our existing linodego code as well as other SDKs like AWS SDK for Go.

instances.go
```diff
@@ -92,6 +93,17 @@ type InstanceBackup struct {
...
+
+var (
```
zliang-akamai, line 102: Can we make them as `const`?

```suggestion
const (
```

instances.go
avestuk: Ok that all works for me, I've rebased with the type being public

instances.go
avestuk: Done

lke_node_pools.go
```diff
@@ -47,7 +47,8 @@ type LKENodePool struct {
 	Linodes []LKENodePoolLinode `json:"nodes"`
 	Tags    []string            `json:"tags"`
 
-	Autoscaler LKENodePoolAutoscaler `json:"autoscaler"`
+	Autoscaler     LKENodePoolAutoscaler  `json:"autoscaler"`
+	DiskEncryption InstanceDiskEncryption `json:"disk_encryption,omitempty"`
```
Michkov, line 51: Based on Epic non-functional requirements it should not be possible to manage DiskEncryption by user. So we should not expose this on LKENodePool.

Michkov (approved):

lgarber-akamai:
@avestuk Thanks for addressing our feedback! 

When you get a chance, could you replace all occurrences of `api.dev.linode.com` with `api.linode.com` in the `TestInstance_DiskEncryption.yaml` fixture? This will fix the failing CI 👍

lgarber-akamai (approved):
LGTM, thanks for the contribution!

jriddle-linode (approved):
Before you merge if you wouldn't mind updating the Title to be a little more descriptive? it will display on the Github Release 
Thanks for the contribution!

k8s/pkg/condition/lke.go
```diff
@@ -33,9 +33,39 @@ func ClusterHasReadyNode(ctx context.Context, options linodego.ClusterConditionO
 	return false, nil
 }
 
+// ClusterHasReadyNode is a ClusterConditionFunc which polls for at least one node to have the
+// condition NodeReady=True.
```
lgarber-akamai, line 37:
```suggestion
// ClusterNodesReady is a ClusterConditionFunc which polls for all nodes to have the
// condition NodeReady=True.
```

avestuk: Doh!

lgarber-akamai (approved):
Looks great, thanks for filling in the gaps!

PR #507, opened 2024-05-14
Expose region capabilities enum

yec-akamai:
## 📝 Description

Add a list of enum for region capabilities. To avoid high-impact breaking change, we only build them into strings instead of a custom type.


lgarber-akamai (approved):
LGTM, nice work!

jriddle-linode (approved):
LGTM!

PR #508, opened 2024-05-21
fix: Prevent unexpected warning from Resty when calling `Client.SetDebug(false)`

lgarber-akamai:
## 📝 Description

This pull request fixes a small issue that would cause the following Resty warning (raised [here](https://github.com/go-resty/resty/blob/v2/client.go#L591)) to be printed when `Client.SetDebug(...)` is called multiple times:

```
WARN RESTY Overwriting an existing on-request-log callback from=github.com/linode/linodego.(*Client).SetDebug.(*Client).sanitizeAuthorizationHeader.func1 to=github.com/linode/linodego.(*Client).SetDebug.(*Client).sanitizeAuthorizationHeader.func1
```

This works by initializing the log hook during client creation rather than when calling `SetDebug(...)`.

## ✔️ How to Test

### Manual Testing

```go
package main

import (
	"github.com/linode/linodego"
)

func main() {
	client.SetDebug(true)
	client.SetDebug(true)
	client.SetDebug(false)
}
```

2. Ensure no warnings are raised from Resty.

### DIFF
--- a/client.go
+++ b/client.go
@@ -122,8 +122,6 @@ func (c *Client) R(ctx context.Context) *resty.Request {
 func (c *Client) SetDebug(debug bool) *Client {
 	c.debug = debug
 	c.resty.SetDebug(debug)
-	// this ensures that if there is an Authorization header present, the value is sanitized/masked
-	c.sanitizeAuthorizationHeader()
 
 	return c
 }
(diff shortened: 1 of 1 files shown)


jriddle-linode (approved):
LGTM

ykim-1 (approved):
LGTM, nice work!

PR #514, opened 2024-05-30
Add new tax_id_invalid event

jcallahan-akamai:
## 📝 Description

This adds the upcoming `tax_id_invalid` Event Action.

This Event Action won't actually be generated by the API until later in June. But, is there any harming in merging now? (If so, we can review/approve, then wait to merge until the API release.)

## ✔️ How to Test

There isn't much to test here, and currently you would need to use devenv to actually generate the event type.


jriddle-linode (approved):
LGTM

PR #523, opened 2024-06-10
doc: Add warning for LKE control plane ACL

lgarber-akamai:
## 📝 Description

This pull request adds notes to the inline function and struct documentation suggesting that LKE ACLs may not be currently available to users.


go.work.sum
```diff
@@ -1,8 +1,46 @@
+cloud.google.com/go/compute/metadata v0.3.0 h1:Tz+eQXMEqDIKRsmY3cHTL6FVaynIjX2QxYC4trgAKZc=
```
lgarber-akamai, line 1: This is unrelated to this change but I went ahead and updated it here 👍

zliang-akamai (approved):

ykim-1 (approved):
LGTM

PR #525, opened 2024-06-12
test: Add Calico Inbound and Outbound policies to LKE nodes for E2E

ykim-1:
## 📝 Description

A new script, `lke_calico_rules_e2e.sh`, has been added and is called in the E2E workflow file after the `Run the integration test suite step` in `e2e-suite.yml`

Key Points:
- `kubectl` and `calicoctl` binaries are required as pre-req which are downloed in the e2e worflow file
- The script retrieves all LKE clusters in the test account and applies Calico rules to the corresponding nodes. If no clusters exist, no action is taken. In most cases, clusters are cleaned and deleted properly after test execution
- Due to the delay in the availability of LKE cluster configurations and the provisioning times of node instances, it is challenging to apply Calico rules directly within the test framework. Therefore, the script is separated and executed independently in the workflow file

## ✔️ How to Test

Forked run - https://github.com/ykim-1/linodego/actions/runs/9491201493/job/26156141578

Forked run on PR - https://github.com/ykim-1/linodego/actions/runs/9491223844/job/26156204794
Note test failed but it was able to apply the calico rules

## 📷 Preview


.github/workflows/ci.yml
```diff
@@ -52,6 +52,14 @@ jobs:
       - name: Install Python deps
         run: pip3 install requests wheel boto3
 
+      - name: Download kubectl and calicoctl for LKE clusters
+        run: |
+          curl -LO "https://storage.googleapis.com/kubernetes-release/release/$(curl -s https://storage.googleapis.com/kubernetes-release/release/stable.txt)/bin/linux/amd64/kubectl"
+          curl -LO "https://github.com/projectcalico/calico/releases/download/v3.25.0/calicoctl-linux-amd64"
+          chmod +x calicoctl-linux-amd64 kubectl
+          mv calicoctl-linux-amd64 /usr/local/bin/calicoctl
+          mv kubectl /usr/local/bin/kubectl
```
zliang-akamai, line 61: Have you tried the Azure's action for installing kubectl?
https://github.com/marketplace/actions/kubectl-tool-installer

.github/workflows/ci.yml
```diff
@@ -64,6 +72,13 @@ jobs:
         env:
           SKIP_LINT: 1
 
+      - name: Apply Calico Rules to LKE
+        if: always()
+        run: |
+          cd scripts && ./lke_calico_rules_e2e.sh
+        env:
+          LINODE_TOKEN: ${{ secrets.DX_LINODE_TOKEN }}
+
```
zliang-akamai, line 81: I thin k the CI never hit the API because it's mocking everything. Is it required for some cases I missed?

ykim-1: Good catch just reverted the ci.yml

zliang-akamai (approved):

PR #535, opened 2024-06-25
Added helper to improve readability of pointer literals

ezilber-akamai:
## 📝 Description

Added helper to improve readability of pointer literals


ykim-1 (approved):
LGTM, tests pass locally

pointer_helpers.go
```diff
@@ -0,0 +1,5 @@
+package linodego
+
+func Pointer[T any](value T) *T {
```
yec-akamai, line 3: Can we have some notes here to explain why and how to use the `Pointer`? Maybe add an example as well.

ezilber-akamai: Good suggestion, thanks!

yec-akamai (approved):
Nice work! 🎉

PR #536, opened 2024-06-27
fix: Create request object per-page in getPaginatedResults(...)

lgarber-akamai:
## 📝 Description

This PR fixes a bug that would cause corrupt/partially overridden data to be returned from endpoint methods that implement `getPaginatedResults[T any](...)` and have a large number of results.

This issue occurred because Resty's request objects are not designed to be executed multiple times. Specifically, the internal `Request.Result` pointer (initialized by us using `SetResult(...)`) is not implicitly reset when executing a request multiple times, leading to new data being written on top of existing data.

### Unit Testing

```
make testunit
```

### Manual Testing

Excluding manual testing steps for now since it requires a large number of resources (~ >200).

### DIFF
--- a/request_helpers.go
+++ b/request_helpers.go
@@ -5,7 +5,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/url"
-	"strconv"
 )
 
 // paginatedResponse represents a single response from a paginated
@@ -30,8 +29,6 @@ func getPaginatedResults[T any](
 
 	result := make([]T, 0)
 
-	req := client.R(ctx).SetResult(resultType)
-
 	if opts == nil {
 		opts = &ListOptions{PageOptions: &PageOptions{Page: 0}}
 	}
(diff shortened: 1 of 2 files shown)


request_helpers_test.go
```diff
@@ -161,6 +165,8 @@ func TestRequestHelpers_delete(t *testing.T) {
 }
 
 func TestRequestHelpers_paginateAll(t *testing.T) {
+	const totalResults = 4123
```
lgarber-akamai, line 168: I made this much larger and uneven to reproduce the failure case before applying the fix

request_helpers.go
```diff
@@ -40,15 +37,20 @@ func getPaginatedResults[T any](
...
 	// Makes a request to a particular page and
 	// appends the response to the result
 	handlePage := func(page int) error {
-		req.SetQueryParam("page", strconv.Itoa(page))
+		// Override the page to be applied in applyListOptionsToRequest(...)
+		opts.Page = page
+
+		// This request object cannot be reused for each page request
+		// because it can lead to possible data corruption
+		req := client.R(ctx).SetResult(resultType)
```
lgarber-akamai, line 48: Another potential solution could've been to call `req.SetResult(...)` at the beginning of this function but I wasn't 100% sure that wouldn't have unintended side-effects

ezilber-akamai (approved):
LGTM!

yec-akamai (approved):
Nice fix! Tests passed locally.

PR #539, opened 2024-07-02
Restored legacy PagedResponse structs and added deprecation notice

ezilber-akamai:
## 📝 Description

Restored legacy PagedResponse structs and added deprecation notice

### DIFF
--- a/paged_response_structs.go
+++ b/paged_response_structs.go
@@ -1,0 +1,55 @@
+package linodego
+
+// Deprecated: AccountAvailabilityPagedResponse exists for historical compatibility and should not be used.
+type AccountAvailabilityPagedResponse legacyPagedResponse[AccountAvailability]
+
+// Deprecated: AccountBetasPagedResponse exists for historical compatibility and should not be used.
+type AccountBetasPagedResponse legacyPagedResponse[AccountBetaProgram]
+
+// Deprecated: BetaProgramPagedResponse exists for historical compatibility and should not be used.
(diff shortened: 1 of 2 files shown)


zliang-akamai:
Just a thought, since they are all in the root package `linodego`, would it be better to put every deprecated structs in a single file?

ezilber-akamai:
> Just a thought, since they are all in the root package `linodego`, would it be better to put every deprecated structs in a single file?

Good idea! It would make it much simpler to phase them out when that eventually happens.

zliang-akamai (approved):
Looks good to me! Nice work!

paged_response_structs.go
```diff
@@ -0,0 +1,55 @@
+package linodego
+
+// Deprecated: AccountAvailabilityPagedResponse exists for historical compatibility and should not be used.
```
lgarber-akamai, line 3: optional: We can do something like this to make this a bit tidier:

```
type (
	// Deprecated: AccountAvailabilityPagedResponse exists for historical compatibility and should not be used.
	AccountAvailabilityPagedResponse legacyPagedResponse[AccountAvailability]

	// Deprecated: AccountBetasPagedResponse exists for historical compatibility and should not be used.
	AccountBetasPagedResponse legacyPagedResponse[AccountBetaProgram]
	
	// ...
)
```

lgarber-akamai (approved):
Looks great, nice work!

PR #540, opened 2024-07-03
Migrate LKE and MySQL Client Functions to Utilize Request Helpers

ezilber-akamai:
## 📝 Description

Migrate LKE and MySQL Client Functions to Utilize Request Helpers


test/integration/lke_clusters_test.go
```diff
@@ -74,8 +74,8 @@ func TestLKECluster_Update(t *testing.T) {
 
 	updatedTags := []string{"test=true"}
 	updatedLabel := cluster.Label + "-updated"
-	updatedK8sVersion := "1.23"
-	isHA := true
+	updatedK8sVersion := "1.30"
+	isHA := false
```
lgarber-akamai, line 78: Note for other reviewers:

As of now HA cannot be specified in the same PUT request as a K8s version upgrade, which caused issues here because we can't guarantee there will be multiple LKE versions available at a time.

lke_clusters.go
```diff
@@ -167,198 +141,138 @@ func (c *Client) ListLKEVersions(ctx context.Context, opts *ListOptions) ([]LKEV
...
-	req := c.R(ctx).SetResult(&LKECluster{})
-	r, err := coupleAPIErrors(req.Get(e))
+	e := formatAPIPath("lke/clusters/%d", clusterID)
+	response, err := doGETRequest[LKECluster](ctx, c, e)
 	if err != nil {
 		return nil, err
 	}
-	return r.Result().(*LKECluster), nil
+
+	return response, nil
```
lgarber-akamai, line 201: optional: This logic can simplify down to `return response, err` 👍

lgarber-akamai (approved):
Looks great, excellent work!

lke_clusters.go
```diff
@@ -167,198 +141,138 @@ func (c *Client) ListLKEVersions(ctx context.Context, opts *ListOptions) ([]LKEV
 		return result.([]LKEVersion), nil
 	}
 
-	err = c.listHelper(ctx, &response, opts)
+	response, err := getPaginatedResults[LKEVersion](ctx, c, "lke/versions", opts)
```
zliang-akamai, line 144: Can we make `"lke/versions"` be in a variable as the SSoT?

zliang-akamai:
This seems much more than lke_cluster_pools and mysql, maybe considering update the title and description to something like: `Migrate LKE and MySQL Client Functions to Utilize Request Helpers`?

zliang-akamai (approved):
Overall looks great!

lke_clusters.go
ezilber-akamai: Fixed!

zliang-akamai (approved):
LGTM!

PR #549, opened 2024-07-10
Add support for LKE NodePool Labels and Taints

mrogers950:
## 📝 Description

LKE recently added "labels" and "taints" parameters to the Node Pool API, so this PR adds support for those parameters. 

## ✔️ How to Test

Running `cd test && make testinit` runs the integration test suite which has two test cases that exercise the functionality, `TestLKENodePool_Update` and `TestLKENodePool_CreateWithLabelsAndTaints`


lke_node_pools.go
```diff
@@ -33,6 +33,16 @@ type LKENodePoolLinode struct {
 	Status     LKELinodeStatus `json:"status"`
 }
 
+// LKENodePoolTaint represents a corev1.Taint to add to an LKENodePool
+type LKENodePoolTaint struct {
+	Key    string `json:"key"`
+	Value  string `json:"value,omitempty"`
+	Effect string `json:"effect"`
```
yec-akamai, line 40: > Effect: The effect value must be NoSchedule, PreferNoSchedule, or NoExecute.

I'm reading the API doc and find the effect value is one of the above. Could you build `Effect` to be an enum type?

yec-akamai (commented):
Overall looks good! Just a small change suggestion. Tests are passing on my end 🎉

lke_node_pools.go
mrogers950: Sure, I've updated this.

yec-akamai (approved):
Nice work, thank you for your contribution!

jriddle-linode (approved):
LGTM, thank you!

lke_node_pools.go
```diff
@@ -41,24 +60,30 @@ type LKENodePool struct {
...
 }
 
 // LKENodePoolUpdateOptions fields are those accepted by UpdateLKENodePoolUpdate
 type LKENodePoolUpdateOptions struct {
-	Count int       `json:"count,omitempty"`
-	Tags  *[]string `json:"tags,omitempty"`
+	Count  int                 `json:"count,omitempty"`
+	Tags   *[]string           `json:"tags,omitempty"`
+	Labels *LKENodePoolLabels  `json:"labels,omitempty"`
+	Taints *[]LKENodePoolTaint `json:"taints,omitempty"`
```
zliang-akamai, line 86: Do you think we make it a simple slice without pointer?
```suggestion
	Taints []LKENodePoolTaint `json:"taints,omitempty"`
```

lgarber-akamai: I think this needs to be a pointer because there is a case where a user wants to set their taints to an explicitly empty list

mrogers950: @lgarber-akamai that's correct, we'll need it to be a pointer.

zliang-akamai (approved):

PR #555, opened 2024-07-19
test: Update getRegionsWithCaps function to also accept a list of plans that must be available in the identified region

ykim-1:
## 📝 Description

A helper function has been updated for the logic to resolve integration test regions that also accepts a list of plans which must be available in the resolved region

Note: the function takes list value of plans as an optional parameter

## ✔️ How to Test

`make testint`

## 📷 Preview


test/integration/integration_suite_test.go
```diff
@@ -191,8 +204,29 @@ func getRegionsWithCaps(t *testing.T, client *linodego.Client, capabilities []st
 		return true
 	}
 
+	// Function to check if a region has the required plans available
+	regionHasPlans := func(regionID string) bool {
```
lgarber-akamai, line 208: Nice use of closures!

yec-akamai (approved):
Tests works well locally, nice work!

test/integration/integration_suite_test.go
```diff
@@ -166,15 +166,28 @@ func transportRecorderWrapper(t *testing.T, fixtureYaml string) (transport.Wrapp
...
+If the plans list is empty, it only checks for the capabilities.
+
+Parameters:
+  - capabilities: A list of required capabilities that regions must support.
+  - plans (optional): A list of required plans that must be available in the regions.
+
+Returns:
+  - string values representing the IDs of regions that meet the given criteria.
+*/
+func getRegionsWithCaps(t *testing.T, client *linodego.Client, capabilities, plans []string) []string {
```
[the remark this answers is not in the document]
ykim-1: This is a great suggestion! I didn't think of putting it into a hash map but this will for sure be faster than my previous solution 🚀

test/integration/integration_suite_test.go
```diff
@@ -191,8 +214,22 @@ func getRegionsWithCaps(t *testing.T, client *linodego.Client, capabilities []st
 		return true
 	}
 
+	// Function to check if a region has the required plans available
+	regionHasPlans := func(regionID string) bool {
+		if len(plans) == 0 {
```
ykim-1, line 219: Still left this check in to avoid any unnecessary map lookups when there are no plans to check.

lgarber-akamai (approved):
Looks great once fixtures are re-run, nice work!

PR #557, opened 2024-07-19
Implemented doRequest helper method

ezilber-akamai:
## 📝 Description

Added a private “doRequest” generic helper that accepts the following parameters:

- The method of the request
- An optional object to use as the body for the request
- An optional pointer to an object to unmarshal the response body to 
- A slice of functions that can be used to mutate the underlying http.Request directly before the request is executed

## ✔️ How to Test

Pull down this PR locally before running tests.

### Unit Tests
`make testunit`


client.go
```diff
@@ -110,6 +152,66 @@ func (c *Client) SetUserAgent(ua string) *Client {
...
+	// Send the request
+	resp, err := c.httpClient.Do(req)
+	if err != nil {
+		return fmt.Errorf("failed to send request: %w", err)
+	}
+	defer resp.Body.Close()
+
+	// Check for HTTP errors
+	if resp.StatusCode < 200 || resp.StatusCode >= 300 {
+		return fmt.Errorf("received non-2xx status code: %d", resp.StatusCode)
```
yec-akamai, line 202: is it possible to also return the detailed reason for the http errors?

yec-akamai (approved):
Works well locally, nice job! Just a small comment.

client.go
ezilber-akamai: Good idea! I will implement this now.

client.go
```diff
@@ -71,6 +74,45 @@ type Client struct {
 	cachedEntryLock *sync.RWMutex
 }
 
+// Client is a wrapper around the Resty client
+type HTTPClient struct {
```
lgarber-akamai, line 78: @ezilber-akamai Any chance we could make this struct private to prevent it from being accessible by outside users? Might also be worth moving this to a separate file for the time being 👍 

Just a bit of context for other reviewers, this is a temporary struct necessary to reduce toe-stepping since we're developing this on the `main` branch

client.go
```diff
@@ -110,6 +152,75 @@ func (c *Client) SetUserAgent(ua string) *Client {
...
+		var errorDetails string
+		if resp.Body != nil {
+			bodyBytes, err := io.ReadAll(resp.Body)
+			if err != nil {
+				errorDetails = fmt.Sprintf("failed to read response body: %v", err)
+			} else {
+				errorDetails = string(bodyBytes)
+			}
+		}
+		return fmt.Errorf("received non-2xx status code: %d, details: %s", resp.StatusCode, errorDetails)
```
lgarber-akamai, line 211: To keep errors unified we might want to build a [`linodego.Error`](https://github.com/linode/linodego/blob/main/errors.go#L24) object here, which may involve creating an equivalent to `coupleAPIErrors` but takes a `*http.Response` rather than a `*resty.Response`.

ezilber-akamai: Thanks for the tip! I will implement this.

client.go

client.go

client.go
```diff
@@ -110,6 +113,67 @@ func (c *Client) SetUserAgent(ua string) *Client {
 	return c
 }
 
+type RequestParams struct {
+	Body     map[string]any
```
lgarber-akamai, line 117: Go's JSON encoder uses reflection to figure out the underlying type of an `any` object, so we can change `Body` to `any` to support directly serializing structs 👍

ezilber-akamai: Thanks for the tip! I fixed this 👍

lgarber-akamai (approved):
Looks perfect, nice work!

PR #562, opened 2024-07-30
Implemented new logging system

ezilber-akamai:
## 📝 Description

Implemented new logging system that no longer relies on resty.

### Unit Tests
`make testunit`

### DIFF
--- a/client.go
+++ b/client.go
@@ -227,6 +227,8 @@ func (c *httpClient) logRequest(req *http.Request, method, url string, bodyBuffe
 	var reqBody string
 	if bodyBuffer != nil {
 		reqBody = bodyBuffer.String()
+	} else {
+		reqBody = "nil"
 	}
 
 	var logBuf bytes.Buffer
--- a/client.go
+++ b/client.go
@@ -127,6 +127,9 @@ func (c *httpClient) doRequest(ctx context.Context, method, url string, params R
 	if params.Body != nil {
 		buf := new(bytes.Buffer)
 		if err := json.NewEncoder(buf).Encode(params.Body); err != nil {
+			if c.debug && c.logger != nil {
+				c.logger.Errorf("failed to encode body: %v", err)
+			}
 			return fmt.Errorf("failed to encode body: %w", err)
 		}
 		bodyReader = buf
@@ -134,6 +137,9 @@ func (c *httpClient) doRequest(ctx context.Context, method, url string, params R
 
 	req, err := http.NewRequestWithContext(ctx, method, url, bodyReader)
 	if err != nil {
+		if c.debug && c.logger != nil {
+			c.logger.Errorf("failed to create request: %v", err)
+		}
 		return fmt.Errorf("failed to create request: %w", err)
 	}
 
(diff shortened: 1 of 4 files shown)


client.go
```diff
@@ -147,26 +153,48 @@ func (c *httpClient) doRequest(ctx context.Context, method, url string, params R
...
+			if c.debug && c.logger != nil {
+				c.logger.Errorf("failed to mutate request: %v", err)
+			}
 			return fmt.Errorf("failed to mutate request: %w", err)
 		}
 	}
 
+	// Log the request if in debug mode
+	if c.debug && c.logger != nil {
+		c.logger.Debugf("sending request: %s %s", method, url)
```
lgarber-akamai, line 165: Could we also show the request body and headers here and in the response log?

ezilber-akamai: Thanks for the feedback! I'll definitely look at text/template.

client.go
```diff
@@ -147,26 +156,89 @@ func (c *httpClient) doRequest(ctx context.Context, method, url string, params R
...
+		if bodyBuffer != nil {
+			reqBody = bodyBuffer.String()
+		}
+
+		reqLog := `Sending request:
+Method: {{.Method}}
+URL: {{.URL}}
+Headers: {{.Headers}}
+Body: {{.Body}}`
+		tmpl, _ := template.New("request").Parse(reqLog)
```
lgarber-akamai, line 178: This looks great, thanks for the quick turnaround!

Since the template itself isn't dynamic, you might be able to move the template creation and parsing to a top-level var so we don't need to re-parse the template every time. This also allows us to validate template errors as soon as the package is loaded using `template.Must(template.New("...")...)`.

client.go
```diff
@@ -118,59 +133,167 @@ type RequestParams struct {
...
 		}
 	}
+	return nil
+}
+
+// nolint:unused
+func (c *httpClient) logRequest(req *http.Request, method, url string, bodyBuffer *bytes.Buffer) {
+	var reqBody string
+	if bodyBuffer != nil {
+		reqBody = bodyBuffer.String()
```
lgarber-akamai, line 229: nitpick: Could we display `nil` if the given body is nil here just to make the output a tiny bit more precise?

lgarber-akamai (approved):
Tested locally and the logs look great!

```2024/08/01 15:33:24.874594 DEBUG RESTY Sending request:
Method: GET
URL: http://127.0.0.1:64109
Headers: map[Accept:[application/json] Content-Type:[application/json]]
Body: 

2024/08/01 15:33:24.875046 DEBUG RESTY Received response:
Status: 200 OK
Headers: map[Content-Length:[21] Content-Type:[text/plain; charset=utf-8] Date:[Thu, 01 Aug 2024 19:33:24 GMT]]
Body: {"message":"success"}
```

client.go
ezilber-akamai: Fixed! Thanks for the tip.

lgarber-akamai (approved):

yec-akamai (approved):
Nice work!

ezilber-akamai:
> Nice work!

Thanks for the review!

PR #566, opened 2024-08-05
Implemented new retry system

ezilber-akamai:
## 📝 Description

Implemented new retry system that no longer relies on resty.

## ✔️ How to Test
`make testunit`


client.go
```diff
@@ -135,45 +135,70 @@ type RequestParams struct {
...
-			return err
+		retryAfter, retryErr := c.retryAfter(resp)
+		if retryErr != nil {
+			return retryErr
 		}
+		time.Sleep(retryAfter)
 	}
 
-	return nil
+	return err
```
lgarber-akamai, line 191: optional: We can use an anonymous function here to reduce nesting a bit. The below snippet is untested but I think it gets the idea across 🙂 

```suggestion
	req, bodyBuffer, err := c.createRequest(ctx, method, url, params)
	if err != nil {
		return err
	}

	if err = c.applyMutators(req, mutators); err != nil {
		return err
	}

	executeRequest := func() (*http.Response, error) {
		if c.debug && c.logger != nil {
			c.logRequest(req, method, url, bodyBuffer)
		}

		resp, err := c.sendRequest(req)
		if err != nil {
			return resp, err
		}

		defer resp.Body.Close()

		if err = c.checkHTTPError(resp); err != nil {
			return resp, err
		}

		if c.debug && c.logger != nil {
			resp, err = c.logResponse(resp)
			if err != nil {
				return resp, err
			}
		}

		return resp, err
	}

	for attempt := 0; attempt < httpDefaultRetryCount; attempt++ {
		resp, err := executeRequest()
		if err != nil {
			return err
		}

		if !c.shouldRetry(resp, err) {
			break
		}

		retryAfter, retryErr := c.retryAfter(resp)
		if retryErr != nil {
			return retryErr
		}

		time.Sleep(retryAfter)
	}
	
	return err
```

client.go
```diff
@@ -135,45 +135,70 @@ type RequestParams struct {
...
-	}
 
-	if params.Response != nil {
-		if err := c.decodeResponseBody(resp, params.Response); err != nil {
-			return err
+		retryAfter, retryErr := c.retryAfter(resp)
+		if retryErr != nil {
+			return retryErr
 		}
+		time.Sleep(retryAfter)
```
lgarber-akamai, line 188: I was a bit confused on this part until I found that 0 is returned by `c.retryAfter(...)` if the `Retry-After` header isn't found. I think that's a good design decision but it might be handy to have a comment clarifying that

client.go
ezilber-akamai: Good idea! This should reduce the complexity a bit.

client.go
ezilber-akamai: Done!

lgarber-akamai (approved):
Looks good and tests are passing locally, great work!

yec-akamai (approved):
Tested and worked well on my end, nice work!

PR #576, opened 2024-09-05
Drop Resty as a dependency

ezilber-akamai:
## 📝 Description

Linodego currently uses the [go-resty package](https://github.com/go-resty/resty) to manage all HTTP requests made to the API. This includes logic for authentication, JSON marshaling/unmarshaling, request retries, and debug outputs.

Unfortunately this package has caused various issues for linodego users, including:

- Various upstream race conditions prevent linodego from being thread-safe
- Uncommitted maintainers lead to difficulty quickly releasing patches for critical issues
- Bloated dependency tree can cause downstream dependency conflicts
- Less flexibility due to limited access to the underlying HTTP client

Because of these issues, Resty is being dropped as a dependency in favor of the mature and widely adopted [net/http package](https://pkg.go.dev/net/http).

## ✔️ How to Test

Run the unit test suite with `make testunit`

Run the integration test suite with `make testint`


lgarber-akamai:
Running the TF cross-test against these changes here: https://github.com/lgarber-akamai/linodego/actions/runs/11111815305/job/30872572090

images.go
```diff
@@ -219,17 +219,22 @@ func (c *Client) CreateImageUpload(ctx context.Context, opts ImageCreateUploadOp
...
 
-	return err
+	resp, err := c.httpClient.Do(req)
+
+	_, err = coupleAPIErrors(resp, err)
+	if err != nil {
+		return err
+	}
+
+	return nil
```
lgarber-akamai, line 237: I noticed this test fails when manually running the fixtures, still trying to pinpoint why 🤔

ezilber-akamai: Interesting, I'll look into this as well. Nice find!

lgarber-akamai: As discussed in Slack, this issue turned out to be caused by the S3 REST API not playing nicely with the Transfer-Encoding used by the net/http package when Content-Length isn't explicitly set to something >= 0.

The test was still failing even after hacking together a fix for that on my end, and it turns out the [test client's transport](https://github.com/ezilber-akamai/linodego/blob/TPT-2965-drop-resty/test/integration/integration_suite_test.go#L161-L164) was appending the `Authorization` header to the upload request despite it not using the `Client{}.createRequest(...)` method. One solution might be to simply delete the authorization header before sending the upload request 🙂

ezilber-akamai: Thanks for digging into this! Your suggestion to override the user-defined roundtripper seems to have fixed the issue.

lgarber-akamai: @ezilber-akamai Thanks for addressing this! Weirdly enough I'm still seeing intermittent errors on the `TestImage_Upload` test, although now they're reporting that the `Content-Length` header is missing. 

It might make sense to use an [io.Seeker](https://stackoverflow.com/a/75459246) to manually discover the buffer size and set the `Content-Length` header accordingly, which will prevent Go from using a chunked transfer encoding 👍

ezilber-akamai: This should be fixed now!

images.go
```diff
@@ -219,17 +219,44 @@ func (c *Client) CreateImageUpload(ctx context.Context, opts ImageCreateUploadOp
...
+			return err
+		}
+
+		_, err = seeker.Seek(0, io.SeekStart)
+		if err != nil {
+			return err
+		}
+
+		contentLength = size
+	}
```
lgarber-akamai, line 239: One more tiny thing, it looks like the client's User-Agent header isn't be respected here. Other than that this looks great!

ezilber-akamai: Good catch! This should be fixed now.

client.go
```diff
@@ -135,27 +158,62 @@ type RequestParams struct {
...
-		req, bodyBuffer, err = c.createRequest(ctx, method, url, params)
+	// Capture the original body into a buffer
+	originalBodyBuffer := new(bytes.Buffer)
+	if params.Body != nil {
+		reader, ok := params.Body.(io.Reader)
+		if !ok {
+			if c.debug && c.logger != nil {
+				c.logger.Errorf("failed to read body: params.Body is not an io.Reader")
+			}
+			return fmt.Errorf("failed to read body: params.Body is not an io.Reader")
```
lgarber-akamai, line 178: optional: We might be able to simplify encapsulate the log and error logic in a helper function.

e.g.
```go
func (c *Client) ErrorAndLogf(format string, args ...string) error { ... }
```

lgarber-akamai:
I think it was implemented in a previous PR, but would it make sense for `RequestParams` to be made private given it's only used by internal helpers?

client.go
```diff
@@ -135,27 +158,62 @@ type RequestParams struct {
...
 		resp       *http.Response
 		err        error
 	)
 
-	for range httpDefaultRetryCount {
-		req, bodyBuffer, err = c.createRequest(ctx, method, url, params)
+	// Capture the original body into a buffer
+	originalBodyBuffer := new(bytes.Buffer)
+	if params.Body != nil {
+		reader, ok := params.Body.(io.Reader)
```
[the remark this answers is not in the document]
ezilber-akamai: Good point. Fixed!

lgarber-akamai (approved):
I tested this pretty thoroughly and everything seems to be working well, great work!

client.go
```diff
@@ -87,6 +102,16 @@ type Client struct {
...
 	cachedEntries   map[string]clientCacheEntry
 	cachedEntryLock *sync.RWMutex
+	logger          Logger
+	requestLog      func(*RequestLog) error
+	onBeforeRequest []func(*http.Request) error
+	onAfterResponse []func(*http.Response) error
+
+	retryConditionals []RetryConditional
+	retryMaxWaitTime  time.Duration
+	retryMinWaitTime  time.Duration
```
zliang-akamai, line 112: Where are `retryMaxWaitTime` and `retryMinWaitTime` applied to the http client? I can't see they are used anywhere..

client.go
```diff
@@ -123,39 +146,64 @@ func init() {
...
-		req, bodyBuffer, err = c.createRequest(ctx, method, url, params)
+	// If there is a body, ensure it's a *bytes.Reader so we can reset its position for each retry
+	var bodyReader *bytes.Reader
+	if params.Body != nil {
+		var ok bool
+		bodyReader, ok = params.Body.(*bytes.Reader)
+		if !ok {
+			return c.ErrorAndLogf("failed to read body: params.Body is not a *bytes.Reader")
+		}
+	}
```
zliang-akamai, line 178: Just curious, if we require `params.Body` to always be `*bytes.Reader`, can we type it as `*bytes.Reader` in the `requestParams` struct?

client.go
```diff
@@ -214,122 +262,131 @@ func (c *httpClient) doRequest(ctx context.Context, method, url string, params R
...
+			return nil, nil, c.ErrorAndLogf("failed to seek to the start of the body: %v", err.Error())
 		}
-		bodyReader = bodyBuffer
+
+		bodyReader = reader
 	}
 
-	req, err := http.NewRequestWithContext(ctx, method, url, bodyReader)
+	req, err := http.NewRequestWithContext(ctx, method, fmt.Sprintf("%s/%s", strings.TrimRight(c.hostURL, "/"),
+		strings.TrimLeft(endpoint, "/")), bodyReader)
```
zliang-akamai, line 296: I think it would be easier to utilize `net/url` package to handle URL concatenation stuff?

client.go
ezilber-akamai: Good catch! These can actually be removed. I initially included them in the Client struct since they were used in the resty client and forgot about them. The http client has it's own timeout property.

ezilber-akamai: Actually, removing them would cause a breaking change that I would like to avoid. I'll keep thinking about this.

client.go
ezilber-akamai: Yes we should be able to, but I typed it as any since that's how we had it before. Do you think it's worth updating the type?

client.go
ezilber-akamai: I tried doing this initially but was running into a lot of issues with the way net/url handled special characters by default.

client.go
ezilber-akamai: Alright this should be fixed now!

zliang-akamai: Thanks @ezilber-akamai!

zliang-akamai: Can you also give them some default values or making them as a pointer type if not already? This can prevent [Go zero values](https://go.dev/tour/basics/12) accidentally being used as the actual configuration.

client.go
zliang-akamai: Since we have already made it as a private struct, I think the risk of changing it is very minimal?

CC @lgarber-akamai (for a second opinion)

https://github.com/linode/linodego/pull/576/files#diff-4b667feae66c9d46b21b9ecc19e8958cf4472d162ce0a47ac3e8386af8bbd8cfL131-R154

ezilber-akamai: Now that you mention it, I'm not sure why I changed it to be private when it was public before. I think at some point I was changing names and accidentally left it lowercase. Do you think we should keep it public like it was before or would it be better to make it private?

client.go
ezilber-akamai: Good call. I added default values in NewClient 👍

client.go
zliang-akamai: I think we are fine to make a such change if we clearly document it in the release note, and I don't think this struct should necessarily be available to other packages.

ezilber-akamai: Alright sounds good! I updated the field to have a type of `*bytes.Reader` instead of `any`.

.github/workflows/ci.yml
```diff
@@ -81,7 +81,7 @@ jobs:
   notify-slack:
     runs-on: ubuntu-latest
     needs: [test]
-    if: always() && github.repository == 'linode/linodego' # Run even if integration tests fail and only on main repository
+    if: always() && github.ref == 'refs/heads/main' && github.event_name == 'push' # Run even if integration tests fail and only on main repository
```
ykim-1, line 84: Do we want to add the condition `github.ref == 'refs/heads/main' && github.event_name == 'push'` along with `github.repository == 'linode/linodego'` to limit slack notifications?

ykim-1 (approved):
LGTM overall, tested locally and also ran tests for terraform. We probably need a follow up ticket to ensure everything runs smooth in terraform and update retry functions. Excellent work!!

.github/workflows/ci.yml
zliang-akamai: I think it's already in `main`? Maybe we can do some main -> proj -> this-branch merges?

client.go
```diff
@@ -206,130 +243,143 @@ func (c *httpClient) doRequest(ctx context.Context, method, url string, params R
...
-func (c *httpClient) applyBeforeRequest(req *http.Request) error {
+func (c *Client) applyBeforeRequest(req *http.Request) error {
 	for _, mutate := range c.onBeforeRequest {
 		if err := mutate(req); err != nil {
-			if c.debug && c.logger != nil {
-				c.logger.Errorf("failed to mutate before request: %v", err)
-			}
-			return fmt.Errorf("failed to mutate before request: %w", err)
+			e := c.ErrorAndLogf("failed to mutate before request: %v", err.Error())
+			return e
```
zliang-akamai, line 315: Something like this can be combined into one line lol
```suggestion
			return c.ErrorAndLogf("failed to mutate before request: %v", err.Error())
```

client.go
```diff
@@ -206,130 +243,143 @@ func (c *httpClient) doRequest(ctx context.Context, method, url string, params R
...
-func (c *httpClient) applyAfterResponse(resp *http.Response) error {
+func (c *Client) applyAfterResponse(resp *http.Response) error {
 	for _, mutate := range c.onAfterResponse {
 		if err := mutate(resp); err != nil {
-			if c.debug && c.logger != nil {
-				c.logger.Errorf("failed to mutate after response: %v", err)
-			}
-			return fmt.Errorf("failed to mutate after response: %w", err)
+			e := c.ErrorAndLogf("failed to mutate after response: %v", err.Error())
+			return e
```
zliang-akamai, line 326: Similar to above, can be reduced to one line

client.go
ezilber-akamai: Oops didn't notice this. Fixed!

client.go

zliang-akamai (approved):
Looks good to me! Thanks for addressing these

PR #582, opened 2024-09-18
Improve image replication test region; Add LA notice

yec-akamai:
## 📝 Description

Improve the image replication test case and remove the hardcoded regions. 

Add LA notice to image gen2 endpoint. 

## ✔️ How to Test

```
make ARGS="-run TestImage_Replicate" fixtures
```

### DIFF
--- a/test/integration/integration_suite_test.go
+++ b/test/integration/integration_suite_test.go
@@ -203,24 +203,8 @@ func getRegionsWithCaps(t *testing.T, client *linodego.Client, capabilities []st
 		t.Fatal(err)
 	}
 
-	regionHasCaps := func(r linodego.Region) bool {
-		capsMap := make(map[string]bool)
-
-		for _, c := range r.Capabilities {
-			capsMap[strings.ToUpper(c)] = true
-		}
-
-		for _, c := range capabilities {
-			if _, ok := capsMap[strings.ToUpper(c)]; !ok {
-				return false
-			}
-		}
-
-		return true
-	}
-
 	for _, region := range regions {
-		if region.Status != "ok" || !regionHasCaps(region) {
+		if region.Status != "ok" || !regionHasCaps(region, capabilities) {
 			continue
 		}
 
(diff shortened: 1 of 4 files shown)


test/integration/integration_suite_test.go
```diff
@@ -270,3 +270,39 @@ func getRegionsWithCapsAndPlans(t *testing.T, client *linodego.Client, capabilit
...
+		}
+
+		for _, c := range capabilities {
+			if _, ok := capsMap[strings.ToUpper(c)]; !ok {
+				return false
+			}
+		}
+
+		return true
+	}
```
lgarber-akamai, line 297: optional: Would it make sense to call `getRegionsWithCaps(...)` here and filter down the results below?

lgarber-akamai (approved):
Image-related tests are all passing on my end, nice work!

test/integration/integration_suite_test.go
yec-akamai: I did try to call `getRegionsWithCaps(...)` at first but figured out it might be better to have the logic here. Because `getRegionsWithCaps(...)` only return the region's id string, we will have to make extra API calls to get the Region object for obtaining necessary info. It's probably slow down the test. 

I'm wondering if it makes sense to move `regionHasCaps(...)` as an independent function, so that we can call it here without repeating this part of logic. What's your thoughts?

ykim-1: I think that would be a good idea. Maybe we can move it as a separate helper function in `integration_test_suite.go`

ykim-1 (approved):
LGTM, tests passed locally with new fixture

test/integration/integration_suite_test.go
yec-akamai: @ykim-1 @lgarber-akamai thanks for the comment! Just applied the improvement, could both of you re-approve the PR?

ykim-1 (approved):

lgarber-akamai (approved):
Looks great, nice work!

PR #583, opened 2024-09-18
Add `last_login` field to `User` struct

zliang-akamai:
## 📝 Description

The field was previously missing, adding it to the struct now.

## ✔️ How to Test
```bash
make fixtures ARGS="-run TestUser"
```

```go
package main

import (
	"context"
	"fmt"
	"log"
	"net/http"
	"os"

	"github.com/linode/linodego"
	"golang.org/x/oauth2"
)

func main() {
	if !ok {
	}

	oauth2Client := &http.Client{
		Transport: &oauth2.Transport{
			Source: tokenSource,
		},
	}

	linodeClient.SetDebug(true)

	usr, _ := linodeClient.GetUser(context.Background(), "YOUR_USERNAME")
	fmt.Println(usr.LastLogin.Status)
}

```

### DIFF
--- a/account_users.go
+++ b/account_users.go
@@ -17,10 +17,17 @@ const (
 	UserTypeDefault UserType = "default"
 )
 
+// LastLogin represents a LastLogin object
+type LastLogin struct {
+	LoginDatetime *time.Time `json:"-"`
+	Status        string     `json:"status"`
+}
+
 // User represents a User object
 type User struct {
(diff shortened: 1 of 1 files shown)


ykim-1 (approved):
LGTM, thanks for the quick follow up 👍

lgarber-akamai (approved):
LGTM!

PR #595, opened 2024-10-25
Updated log formatting to more closely resemble Resty's

ezilber-akamai:
## 📝 Description

Improved the formatting of logs generated with the new logging system. 

## ✔️ How to Test

After pulling down this branch, run `make testunit` and `make testint`.

Play around with the new logging system locally to see the new formatting.


client.go
```diff
@@ -51,30 +51,41 @@ const (
...
-URL: {{.URL}}
-Headers: {{.Headers}}
-Body: {{.Body}}`))
-
-	respLogTemplate = template.Must(template.New("response").Parse(`Received response:
-Status: {{.Status}}
-Headers: {{.Headers}}
-Body: {{.Body}}`))
+	reqLogTemplate = template.Must(template.New("request").Parse(`
+============================================================================================
```
lgarber-akamai, line 55: Optional: We can use `go:embed` here to avoid hardcoding the template in a string literal. See: https://pkg.go.dev/embed

ezilber-akamai: Done! Thanks for the tip 😃

jcallahan-akamai (commented):
Nice work, this mostly looks great. I tested manually and found one bug and one potential enhancement.

**Bug**: it doesn't seem like the `BODY` of POST or PUT requests are being logged. To reproduce, try a simple request, like: 

```go
instanceOpts := linodego.InstanceUpdateOptions{
	Label: "new-label",
}

instance, err := linodeClient.UpdateInstance(context.Background(), linodeID, instanceOpts)
if err != nil {
	log.Fatalf("failed to update instance: %v", err)
}
```

Resulting request logged:
```
.....
HEADERS:
    User-Agent: linodego/v1.24.0 https://github.com/linode/linodego
    Authorization: Bearer *******************************
    Content-Type: application/json
    Accept: application/json
BODY:
--------------------------------------
```

**Enhancement**: it would be great to sort HTTP headers alphabetically, for readability.

ezilber-akamai:
> Nice work, this mostly looks great. I tested manually and found one bug and one potential enhancement.
> 
> **Bug**: it doesn't seem like the `BODY` of POST or PUT requests are being logged. To reproduce, try a simple PUT/POST, like:
> 
> ```go
> opts := linodego.InstanceUpdateOptions{
> 	Label: "new-label",
> }
> instance, err := linodeClient.UpdateInstance(context.Background(), linodeID, opts)
> ```
> 
> Resulting request logged:
> 
> ```
> .....
> HEADERS:
>     User-Agent: linodego/v1.24.0 https://github.com/linode/linodego
>     Authorization: Bearer *******************************
>     Content-Type: application/json
>     Accept: application/json
> BODY:
> --------------------------------------
> ```
> 
> **Enhancement**: it would be great to sort HTTP headers alphabetically, for readability.

Thanks for the review! Good catch on the issue with request bodies not being logged. I think I know what the issue is and it should be a fairly simple fix. Good idea about sorting the HTTP headers. I will implement this as well.

ezilber-akamai:
> Nice work, this mostly looks great. I tested manually and found one bug and one potential enhancement.
> 
> **Bug**: it doesn't seem like the `BODY` of POST or PUT requests are being logged. To reproduce, try a simple PUT/POST, like:
> 
> ```go
> opts := linodego.InstanceUpdateOptions{
> 	Label: "new-label",
> }
> instance, err := linodeClient.UpdateInstance(context.Background(), linodeID, opts)
> ```
> 
> Resulting request logged:
> 
> ```
> .....
> HEADERS:
>     User-Agent: linodego/v1.24.0 https://github.com/linode/linodego
>     Authorization: Bearer *******************************
>     Content-Type: application/json
>     Accept: application/json
> BODY:
> --------------------------------------
> ```
> 
> **Enhancement**: it would be great to sort HTTP headers alphabetically, for readability.

Fixed the bug and added alphabetic sorting of headers!

lgarber-akamai (commented):
This looks just about perfect, great work! 

I'm not sure if this is directly related to this PR but it looks like the `Authorization` header sanitization logic [here](https://github.com/linode/linodego/blob/6f5809c250bb7582176a840e4f7d1af1e0a86706/client.go#L822) is causing the header to be overwritten before it makes it to the API, causing a `[401] Invalid Token` error.

lgarber-akamai (approved):
Everything is working well when testing this manually and in the Terraform provider. I also diffed between the outputs of the legacy and new logging systems and everything matches perfectly. Excellent work!

zliang-akamai (approved):
LGTM!

PR #615, opened 2024-11-12
Disable `recvcheck` linter, remove EOL linters

zliang-akamai:
## 📝 Description

[`recvcheck` requires receiver type of each method of the same struct to be the same](https://github.com/raeperd/recvcheck#motivtation), which is not realistic for our SDK.

Also removed some EOL linters.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -23,7 +23,6 @@ linters:
   enable-all: true
   disable:
     # deprecated linters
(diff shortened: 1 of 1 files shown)


lgarber-akamai (approved):
Great work!

yec-akamai (approved):
Thanks for cleaning up the linters!

PR #634, opened 2024-11-26
Added support for missing Obj-related endpoints

ezilber-akamai:
## 📝 Description

Added support for the `GET object-storage/buckets/$1/$2/object-list` endpoint. Also reran some fixtures that were not properly sanitized.

## ✔️ How to Test

`make testunit`
`make testint`


yec-akamai (approved):
Tests passed locally!

zliang-akamai:
Can we use Google's `go-querystring` library to simplified query string generation? I think this would be much safer than implementing it in-house.

ezilber-akamai:
> Can we use [Google's `go-querystring`](https://github.com/google/go-querystring) library to simplified query string generation? I think this would be much safer than implementing it in-house.

Good idea! This is fixed now.

zliang-akamai (approved):
Looks good and works well to me, nice job!

PR #639, opened 2024-12-05
Fix: Updated PG migrations field to be pointer

ezilber-akamai:
## 📝 Description

Changed `migrations` field of PlacementGroup to be a pointer to fit the API contract. Also, updated several instances that were being created with an image of `linode/debian9`, which is no longer supported.

## ✔️ How to Test

`make testint`
`make testunit`


lgarber-akamai (approved):
Looks great, thanks for fixing this!

yec-akamai (approved):
LGTM. Thanks for the fix!

PR #641, opened 2024-12-11
new: Introduce UpdateIPAddressV2; deprecate UpdateIPAddress

lgarber-akamai:
## 📝 Description

This pull request introduces a new `UpdateIPAddressV2(...)` endpoint function and marks the `UpdateIPAddress(...)` endpoint method as deprecated. This is necessary because a nil value for RDNS is a valid case (resetting the RDNS address to default) but could no longer be done with the addition of `omitempty`, which was required when adding the Reserved field.

## ✔️ How to Test

### Integration Testing

```
make ARGS="-run TestIPAddress" fixtures
```

### Manual Testing

**NOTE: IP reservation is _not_ covered by these testing steps.**

```go
package main

import (
	"context"
	"fmt"
	"log"
	"os"

	"github.com/linode/linodego"
)

func main() {

	inst, err := client.CreateInstance(context.Background(), linodego.InstanceCreateOptions{
		Region: "us-mia",
		Type:   "g6-nanode-1",
		Label:  "foobar",
	})
	if err != nil {
		log.Fatal(err)
	}

	// Update RDNS to nip.io
	ip, err := client.UpdateIPAddressV2(
		context.Background(),
		inst.IPv4[0].String(),
		linodego.IPAddressUpdateOptionsV2{
			RDNS: linodego.Pointer(linodego.Pointer(fmt.Sprintf("%s.nip.io", inst.IPv4[0].String()))),
		},
	)
	if err != nil {
		log.Fatal(err)
	}

	fmt.Println("Post-Update RDNS:", ip.RDNS)

	// Update RDNS to default
	ip, err = client.UpdateIPAddressV2(
		context.Background(),
		inst.IPv4[0].String(),
		linodego.IPAddressUpdateOptionsV2{
			RDNS: linodego.Pointer[*string](nil),
		},
	)
	if err != nil {
		log.Fatal(err)
	}

	fmt.Println("Post-Reset RDNS:", ip.RDNS)
}
```

2. Ensure the output looks similar to the following:

```
Post-Update RDNS: 172.233.171.197.nip.io
Post-Reset RDNS: 172-233-171-197.ip.linodeusercontent.com
```

### DIFF
--- a/test/unit/account_agreements_test.go
+++ b/test/unit/account_agreements_test.go
@@ -2,10 +2,11 @@ package unit
 
 import (
 	"context"
+	"testing"
+
 	"github.com/jarcoal/httpmock"
 	"github.com/linode/linodego"
 	"github.com/stretchr/testify/assert"
-	"testing"
 )
 
 func TestAccountAgreements_Get(t *testing.T) {
--- a/network_ips.go
+++ b/network_ips.go
@@ -4,11 +4,22 @@ import (
 	"context"
 )
 
-// IPAddressUpdateOptions fields are those accepted by UpdateToken
-type IPAddressUpdateOptions struct {
+// IPAddressUpdateOptionsV2 fields are those accepted by UpdateIPAddress.
+// NOTE: An IP's RDNS can be reset to default using the following pattern:
+//
+//	IPAddressUpdateOptionsV2{
+//		RDNS: linodego.Pointer[*string](nil),
+//	}
+type IPAddressUpdateOptionsV2 struct {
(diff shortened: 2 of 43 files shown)


test/unit/account_agreements_test.go
```diff
@@ -2,10 +2,11 @@ package unit
 
 import (
 	"context"
+	"testing"
```
lgarber-akamai, line 5: These were all changed automatically by `gofumpt`.

ezilber-akamai (approved):
Local integration tests are passing and manual testing is working as expected. Nice work!

PR #645, opened 2024-12-17
VPU Support

jriddle-linode:
## 📝 Description

Adds `AcceleratedDevices` to `Instnace.Stats` and `GET linode/types`
Also adds `NETINT Quadra T1U` Capability enum.

## ✔️ How to Test

```
make fixtures ARGS="-run TestInstance_withVPU"
```

### DIFF
--- a/regions.go
+++ b/regions.go
@@ -22,6 +22,7 @@ const (
 	CapabilityVlans                  string = "Vlans"
 	CapabilityVPCs                   string = "VPCs"
 	CapabilityVPCsExtra              string = "VPCs Extra"
+	CapabilityVPU                    string = "NETINT Quadra T1U"
 	CapabilityMachineImages          string = "Machine Images"
 	CapabilityBareMetal              string = "Bare Metal"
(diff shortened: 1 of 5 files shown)


regions.go
```diff
@@ -22,6 +22,7 @@ const (
 	CapabilityVlans                  string = "Vlans"
 	CapabilityVPCs                   string = "VPCs"
 	CapabilityVPCsExtra              string = "VPCs Extra"
+	CapabilityVPU                    string = "NETINT Quadra T1U"
```
zliang-akamai, line 25: Should we consider naming it as `CapabilityQuadraT1UVPU` just in case there will be more VPU models in the future?

zliang-akamai (approved):
Tests passed, looks good to me!

regions.go
jriddle-linode: I dont think that will impact capability but to be safe lets do that.

ykim-akamai (approved):
LGTM, tests pass locally. Nice work!

PR #654, opened 2025-01-10
RFC: `doPOSTRequestNoResponseBody` Request Helper that doesn't Expecets Return Value

zliang-akamai:
## 📝 Description
This may simplify some request helper call to `doPOSTRequest`. Examples are included in files changed.

This is a proposal for discussion, feel free to suggest change or objection of merging it.


ezilber-akamai (approved):
I think this is a good addition! Definitely simplifies quite a few endpoint functions so I see no reason not to use this new helper.

lgarber-akamai (approved):
Great idea and nice work!

PR #635, opened 2024-12-02
NewClient panics if http.client is nil and LINODE_CA is set

kokes:
I tried a new linodego with a custom CA and a nil HTTP client and encountered a panic. Here's a minimal reproducer (that I now include in the test suite):

```
func TestClient_CustomRootCAWithoutCustomRoundTripper(t *testing.T) {
	caFile, err := os.CreateTemp(t.TempDir(), "linodego_test_ca_*")
	if err != nil {
		t.Fatalf("Failed to create temp ca file: %s", err)
	}
	defer os.Remove(caFile.Name())

	t.Setenv(APIHostCert, caFile.Name())

        NewClient(nil)
}
```

Once I run this test, I get:

```
$ go test -run=TestClient_CustomRootCAWithoutCustomRoundTripper ./...
?   	github.com/linode/linodego/cmd	[no test files]
?   	github.com/linode/linodego/internal/parseabletime	[no test files]
?   	github.com/linode/linodego/internal/testutil	[no test files]
--- FAIL: TestClient_CustomRootCAWithoutCustomRoundTripper (0.00s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x0 pc=0x102dbb3d8]

goroutine 22 [running]:
testing.tRunner.func1.2({0x102fb5dc0, 0x1034115e0})
	/opt/homebrew/Cellar/go/1.23.3/libexec/src/testing/testing.go:1632 +0x1bc
testing.tRunner.func1()
	/opt/homebrew/Cellar/go/1.23.3/libexec/src/testing/testing.go:1635 +0x334
panic({0x102fb5dc0?, 0x1034115e0?})
	/opt/homebrew/Cellar/go/1.23.3/libexec/src/runtime/panic.go:785 +0x124
github.com/linode/linodego.NewClient(0x0)
	/Users/okokes/git/linode/public/linodego/client.go:741 +0x208
github.com/linode/linodego.TestClient_CustomRootCAWithoutCustomRoundTripper(0x140000c6b60)
	/Users/okokes/git/linode/public/linodego/client_test.go:589 +0xdc
testing.tRunner(0x140000c6b60, 0x10306a230)
	/opt/homebrew/Cellar/go/1.23.3/libexec/src/testing/testing.go:1690 +0xe4
created by testing.(*T).Run in goroutine 1
[...]

PR #663, opened 2025-01-28
Checking for DefaultTransport in NewClient

kokes:
An earlier PR - #593 (later adjusted in #635) - introduced a new check for `client.Transport`, but it turns out it does not behave as expected. While our tests cover the cases of `http.Client == nil` and `http.Client{Transport: custom}`, they don't cover the case of `http.Client{Timeout: foo}` (or any other non-transport setting). In such a case, `Transport` is set to `DefaultTransport`, BUT this value is not comparable to `http.DefaultTransport`.

~~I've added new tests to uncover this case, but I have yet to develop a correct implementation of this logic.~~

My solution is to not compare against `http.DefaultTransport`. We just check if the transport is the upstream `http.Transport` and if not, we assume this was reimplemented and thus its TLS stack won't be reconfigurable by Resty (see https://github.com/go-resty/resty/blob/1e19d6bc6a102df59e1997a09850d0c0532e7f9a/client.go#L860-L864 - this is where Resty fails if `client.Transport` is not `http.Transport` - but it only logs, no errors are propagated :sigh:)

I tested linodego with and without this PR and it has fixed an issue for us when we have `LINODE_CA=... && http.Client == http.Client{Timeout: foo}` (also one of the new test cases).

tl;dr: `client.Transport == http.DefaultTransport` is always `false`, this comparison has never worked, so any non-nil `http.Client` will fail this and trigger incorrect handling of `LINODE_CA`

### DIFF
--- a/client.go
+++ b/client.go
@@ -882,7 +882,7 @@ func generateListCacheURL(endpoint string, opts *ListOptions) (string, error) {
 }
 
 func hasCustomTransport(hc *http.Client) bool {
-	if hc == nil {
+	if hc == nil || hc.Transport == nil {
 		return false
 	}
 	if hc.Transport != http.DefaultTransport.(*http.Transport) {
--- a/client.go
+++ b/client.go
@@ -885,7 +885,7 @@ func hasCustomTransport(hc *http.Client) bool {
 	if hc == nil || hc.Transport == nil {
 		return false
 	}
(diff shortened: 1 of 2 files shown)


lgarber-akamai (approved):
This change looks good to me, thank you for the contribution!

shanduur-akamai (approved):

zliang-akamai (approved):

PR #669, opened 2025-02-04
add support for the type field when creating nodebalancers

eljohnson92:
## 📝 Description

This PR adds support for a new `type` of nodebalancer called `premium`
## ✔️ How to Test

This can be tested by sending a request for an explicit `common` or `premium` nodebalancer type

### DIFF
--- a/nodebalancer.go
+++ b/nodebalancer.go
@@ -26,6 +26,8 @@ type NodeBalancer struct {
 	ClientConnThrottle int `json:"client_conn_throttle"`
 	// Information about the amount of transfer this NodeBalancer has had so far this month.
 	Transfer NodeBalancerTransfer `json:"transfer"`
+	// This NodeBalancer's plan Type
+	Type NodeBalancerPlanType `json:"type"`
 
 	// An array of tags applied to this object. Tags are for organizational purposes only.
 	Tags []string `json:"tags"`
@@ -58,6 +60,7 @@ type NodeBalancerCreateOptions struct {
 	Configs            []*NodeBalancerConfigCreateOptions `json:"configs,omitempty"`
 	Tags               []string                           `json:"tags"`
 	FirewallID         int                                `json:"firewall_id,omitempty"`
+	Type               NodeBalancerPlanType               `json:"type,omitempty"`
 	VPCs               []NodeBalancerVPCOptions           `json:"vpcs,omitempty"`
 }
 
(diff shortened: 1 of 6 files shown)


nodebalancer.go
```diff
@@ -26,6 +26,8 @@ type NodeBalancer struct {
 	ClientConnThrottle int `json:"client_conn_throttle"`
 	// Information about the amount of transfer this NodeBalancer has had so far this month.
 	Transfer NodeBalancerTransfer `json:"transfer"`
+	// This NodeBalancer's plane Type
```
rahulait, line 29:
```suggestion
	// This NodeBalancer's plan type
```

yec-akamai (commented):
@eljohnson92 Thank you for the contribution! Can you add test cases, i.e. integration test, to test through the new field?

There are examples you can find here: https://github.com/linode/linodego/tree/main/test/integration
And you can test and generate the fixtures by running `make TEST_ARGS="-run $TestCaseName" fixtures`

eljohnson92:
@yec-akamai I can definitely add this, the only issue is currently the Linode API does not return what the type is from a created NB so we can test that submission works but not validate it is the right type, do you think just submission is good for now?

test/integration/nodebalancers_test.go
```diff
@@ -13,8 +13,29 @@ var (
...
+	// when comparing fixtures to random value Label will differ, compare the known suffix
+	if !strings.Contains(*nodebalancer.Label, label) {
+		t.Errorf("nodebalancer returned does not match nodebalancer create request")
+	}
+
+	assertDateSet(t, nodebalancer.Created)
+	assertDateSet(t, nodebalancer.Updated)
+}
+
+func TestNodeBalancer_Create_Type(t *testing.T) {
```
yec-akamai, line 37: Can you assert the value of `nodebalancer.type` in this test?

test/integration/nodebalancers_test.go
```diff
@@ -80,7 +101,7 @@ func TestNodeBalancer_Get(t *testing.T) {
 	}
 }
 
-func setupNodeBalancer(t *testing.T, fixturesYaml string) (*linodego.Client, *linodego.NodeBalancer, func(), error) {
+func setupNodeBalancer(t *testing.T, fixturesYaml string, testOpts NBTestOptions) (*linodego.Client, *linodego.NodeBalancer, func(), error) {
```
yec-akamai, line 104: Thank you for modifying the helper function for better usage when creating nodebalancer. It might be better if we can take in a list of createOps modifiers to allow flexibility of changing the arguments. Similar as: https://github.com/linode/linodego/blob/main/test/integration/lke_clusters_test.go#L318

yec-akamai (approved):
Tests all passed on my end, nice work!

ykim-akamai (approved):
LGTM, thanks for the contribution

PR #689, opened 2025-02-25
Nodebalancer Config Changes for VPC integration

komer3:
## 📝 Description

This is a follow up PR to the previous one where we added changes to enable creating a Nodebalancers in VPC. This PR adds the functionality to allow configuring the Nodebalancer and Nodes in VPC

## ✔️ How to Test

## 📷 Preview


nodebalancer_config_nodes.go
```diff
@@ -14,6 +14,7 @@ type NodeBalancerNode struct {
 	Mode           NodeMode `json:"mode"`
 	ConfigID       int      `json:"config_id"`
 	NodeBalancerID int      `json:"nodebalancer_id"`
+	VPCConfigID    int      `json:"vpc_config_id,omitempty"`
```
lgarber-akamai, line 17: I think the `omitempty` can be dropped here since `NodeBalancerNode` isn't expected to be serialized into an API request body.

lgarber-akamai (approved):
Fixtures are passing on my end, thanks for the contribution!

nodebalancer_config_nodes.go
komer3: Updated!

zliang-akamai (approved):

PR #701, opened 2025-03-24
Support new fields for LKE-E node pool

yec-akamai:
## 📝 Description

Add support to create and update LKE-E node pool with new fields `K8sVersion`, `UpdateStrategy`. The new fields allow to upgrade nodes version under a specific node pool, and describe the upgrade strategy to use. They're for LKE enterprise cluster only. 

## ✔️ How to Test

Unit test:
```
make test-unit
```

Integration test:
```
make test-int
```

### DIFF
--- a/test/integration/lke_node_pools_test.go
+++ b/test/integration/lke_node_pools_test.go
@@ -312,3 +312,77 @@ func setupLKENodePool(t *testing.T, fixturesYaml string, nodePoolCreateOpts *lin
 	}
 	return client, lkeCluster, pool, teardown, err
 }
+
+func TestLKEEnterpriseNodePoolK8sUpgrade_Get(t *testing.T) {
+	testLKENodePoolCreateOpts.K8sVersion = linodego.Pointer("v1.31.1+lke1")
+	testLKENodePoolCreateOpts.UpdateStrategy = linodego.Pointer(linodego.LKENodePoolRollingUpdate)
+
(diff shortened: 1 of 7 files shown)


ezilber-akamai (approved):
Tests passing locally!

test/integration/lke_node_pools_test.go
```diff
@@ -312,3 +312,76 @@ func setupLKENodePool(t *testing.T, fixturesYaml string, nodePoolCreateOpts *lin
...
+	}
+}
+
+func setupLKEEnterpriseNodePool(t *testing.T, fixturesYaml string, nodePoolCreateOpts *linodego.LKENodePoolCreateOptions) (*linodego.Client, *linodego.LKECluster, *linodego.LKENodePool, func(), error) {
+	t.Helper()
+	var fixtureTeardown func()
+	client, lkeCluster, fixtureTeardown, err := setupLKECluster(t, []clusterModifier{func(createOpts *linodego.LKEClusterCreateOptions) {
+		createOpts.Tier = "enterprise"
+		createOpts.Region = "us-lax"
+		createOpts.K8sVersion = "v1.31.1+lke1"
```
ezilber-akamai, line 368: This should probably be updated to dynamically fetch an enterprise version

jriddle-linode (approved):
Looks good and works locally

test/integration/lke_node_pools_test.go
yec-akamai: I think in linodego testing it's probably fine because we use the fixtures

PR #707, opened 2025-04-02
Update k8sversion in LKE-E test

ykim-akamai:
## 📝 Description

Addressing smoke test failure due to outdated k8sversion for LKE E 

```
    lke_clusters_test.go:68: failed to create LKE cluster: [400] [k8s_version] k8s_version is not valid
```
## ✔️ How to Test

`make fixtures TEST_ARGS="-run TestLKECluster_Enterprise_smoke"`

## 📷 Preview


ezilber-akamai (approved):
Thanks for the fix! The test is passing locally for me. Perhaps it could be worth fetching the version via the API so that if it changes again in the future, the test will still pass?

PR #714, opened 2025-04-07
Added support for Configurable DB Params

ezilber-akamai:
## 📝 Description

Implemented support for Configurable DB Params and added unit tests using mocks to test the new features.

## ✔️ How to Test
`make test-unit`


mysql.go
```diff
@@ -50,6 +50,374 @@ type MySQLDatabase struct {
 	UsedDiskSizeGB    int                       `json:"used_disk_size_gb"`
 	TotalDiskSizeGB   int                       `json:"total_disk_size_gb"`
 	Port              int                       `json:"port"`
+
+	EngineConfig MySQLDatabaseEngineConfig `json:"engine_config"`
```
yec-akamai, line 54: Should we use `MySQLDatabaseConfigInfo` because it looks only for read here? Similarly for the `PostgresDatabase` below.

mysql.go
```diff
@@ -50,6 +50,374 @@ type MySQLDatabase struct {
...
+}
+
+type MySQLDatabaseEngineConfig struct {
```
yec-akamai, line 57: I wonder if it's better to add an `options` suffix, like `MySQLEngineConfigOptions`, since it's for the update options.

yec-akamai (commented):
Overall looks good! Might be better to add a comment to each new struct to describe their definition and usage.

mysql.go
ezilber-akamai: The `Info` structs are used for the new `GET /databases/mysql/config` and `GET /databases/postgresql/config` endpoints and have different fields the `MySQLDatabaseEngineConfig` struct, which only contains the value for each config rather than info about it.

mysql.go
ezilber-akamai: Right now, I have the `MySQLDatabaseEngineConfig` struct being used in the `GET /databases/mysql/instances/{instanceId}`, `POST /databases/mysql/instances/{instanceId}`, and `PUT /databases/mysql/instances/{instanceId}` endpoints, and I assume that these endpoints will always return the same sub-fields for the `engine_config` field. Since it is also returned in the GET endpoint, I think the `Options` suffix might be confusing.

That being said, the naming convention I went with is definitely confusing, so if you have any better ideas I'm happy to change them.

yec-akamai: Oh I misread the fields under these two and thought they serve for diverging the read and write path. In this case I think we're good here. Sorry for the confusion.

yec-akamai (approved):
Unit tests passed locally, nice work!

mysql.go
ezilber-akamai: No worries!

ykim-akamai (approved):
LGTM, thank you for writing the unit tests so thoroughly 💯

PR #719, opened 2025-04-15
Fix lint

zliang-akamai:
## 📝 Description

This is to fix the lint by grouping unexported and exported functions.

## ✔️ How to Test

```bash
golangci-lint run
```


lgarber-akamai (approved):
Thank you for addressing this!

PR #720, opened 2025-04-15
Added support for Host/VM Maintenance

ezilber-akamai:
## 📝 Description

Implemented changes for Host/VM Maintenance Policy and added unit tests.

## ✔️ How to Test

`make test-unit`


account_events.go
```diff
@@ -32,9 +31,6 @@ type Event struct {
 	// If this Event has been seen.
 	Seen bool `json:"seen"`
 
-	// The estimated time remaining until the completion of this Event. This value is only returned for in-progress events.
-	TimeRemaining *int `json:"-"`
```
lgarber-akamai, line 36: Just curious, was this field removed from the API response? Ideally we'd probably want to mark it as a deprecated to prevent a breaking change since I can't really recall whether this field ever returned valuable info

ezilber-akamai: Unclear. In the API spec, the field was in the `before` response and not in the `after` response. It could just be a mistake, but yeah I can just mark it as deprecated for the time being and if the actual API response still returns the field then we can just drop the deprecation notice.

jriddle-linode (approved):
LGTM and works locally

PR #724, opened 2025-04-17
Adding VPC Enhanced Interfaces

jriddle-linode:
## 📝 Description

Initial additions for interfaces

## ✔️ How to Test

```bash
make test-unit
```


interfaces.go
```diff
@@ -0,0 +1,171 @@
...
+	DefaultRoute InterfacesDefaultRoute `json:"default_route"`
+	Version      int                    `json:"version"`
+	VPC          InterfacesVPC          `json:"vpc"`
+	Public       InterfacesPublic       `json:"public"`
+	VLAN         InterfacesVLAN         `json:"vlan"`
+}
+
+type InterfacesDefaultRoute struct {
+	IPv4 bool `json:"ipv4,omitempty"`
+	IPv6 bool `json:"ipv6,omitempty"`
```
zliang-akamai, line 25:
```suggestion
	IPv4 *bool `json:"ipv4,omitempty"`
	IPv6 *bool `json:"ipv6,omitempty"`
```
Let's make them pointers to avoid confusions

interfaces.go
```diff
@@ -0,0 +1,171 @@
...
+}
+
+type InterfacesVPC struct {
+	SubnetID int             `json:"subnet_id"`
+	IPv4     *InterfacesIPv4 `json:"ipv4,omitempty"`
+}
```
zliang-akamai, line 31:
```suggestion
type InterfacesVPC struct {
	VPCID    int `json:"vpc_id"`
	SubnetID int             `json:"subnet_id"`
	IPv4     *InterfacesIPv4 `json:"ipv4,omitempty"`
}
```
VPC ID is missing here

zliang-akamai: Should we separate return and post objects for these interfaces? `vpc_id` will only appear in the returning object...

interfaces.go
```diff
@@ -0,0 +1,171 @@
...
+
+type InterfacesVLAN struct {
+	Label       string  `json:"vlan_label"`
+	IPAMAddress *string `json:"ipam_address,omitempty"`
+}
+
+type InterfacesIPv4 struct {
+	Addresses []*InterfacesAddress `json:"addresses,omitempty"`
+}
```
zliang-akamai, line 45: The return object also contains `shared` object in the IPv4 object for public interfaces, and `ranges` for VPC interfaces. Maybe we can separate the ipv4 struct for each type of interface?

interfaces.go
```diff
@@ -0,0 +1,171 @@
...
+
+type InterfacesIPv6 struct {
+	Addresses []*InterfacesAddress `json:"addresses,omitempty"`
+	Ranges    []*InterfacesRange   `json:"ranges,omitempty"`
+}
```
zliang-akamai, line 50: `slaac` and `shared` list may be in the return object, but `addresses` won't be in the return object...

interfaces.go
```diff
@@ -0,0 +1,205 @@
...
+
+type InterfacesIPv4 struct {
+	Addresses []*InterfacesIPv4Address `json:"addresses,omitempty"`
+	Shared    []*InterfacesIPv4Shared  `json:"shared,omitempty"`
+}
+
+type InterfacesIPv6 struct {
+	Ranges []*InterfacesIPv6Range `json:"ranges,omitempty"`
+	Shared []*InterfacesIPv6Range `json:"shared,omitempty"`
```
jriddle-linode, line 51: These both have the same shape so i used the same object, if this is confusing i could separate them

zliang-akamai: Should we use `[]InterfacesIPv6Range` rather than `[]*InterfacesIPv6Range`? I think there is not a list of null for this field?

interfaces.go
```diff
@@ -0,0 +1,205 @@
...
+
+type InterfacesCreateVLAN struct {
+	IpamAddress string `json:"ipam_address"`
+	VLANLabel   string `json:"vlan_label"`
+}
+
+type InterfacesCreateVPC struct {
+	IPv4     []*InterfacesIPv4 `json:"ipv4,omitempty"`
+	SubnetID int               `json:"subnet_id"`
+}
```
zliang-akamai, line 104: Should we rename them into `XXXInterfaceCreateOptions` to be consistent with other naming in the repo?

interfaces.go
jriddle-linode: good catch, updated several others based on the spec

interfaces.go
jriddle-linode: Updated, let me know if you would like any changes, i tried to reuse the get structs where i could but if thats confusing we can just dupe

interfaces.go
```diff
@@ -0,0 +1,215 @@
...
+	Ranges []InterfacesIPv6RangeCreateOptions `json:"ranges"`
+}
+
+type InterfacesIPv6RangeCreateOptions struct {
+	Range string `json:"range"`
+}
+
+type InterfacesVPCCreateOptions struct {
+	SubnetID int               `json:"subnet_id"`
+	IPv4     []*InterfacesIPv4 `json:"ipv4,omitempty"`
```
ChihweiLHBird, line 113:
```suggestion
	IPv4     []InterfacesIPv4 `json:"ipv4,omitempty"`
```

interfaces.go
```diff
@@ -0,0 +1,215 @@
...
+	FirewallID   *int                           `json:"firewall_id,omitempty"`
+	DefaultRoute *InterfacesDefaultRoute        `json:"default_route,omitempty"`
+	Public       *InterfacesPublicCreateOptions `json:"public,omitempty"`
+	VPC          *InterfacesVPCCreateOptions    `json:"vpc,omitempty"`
+	VLAN         *InterfacesVLAN                `json:"vlan,omitempty"`
+}
+
+type InterfacesPublicCreateOptions struct {
+	IPv4 []*InterfacesIPv4              `json:"ipv4,omitempty"`
+	IPv6 []*InterfacesIPv6CreateOptions `json:"ipv6,omitempty"`
```
ChihweiLHBird, line 100:
```suggestion
	IPv4 []InterfacesIPv4              `json:"ipv4,omitempty"`
	IPv6 []InterfacesIPv6CreateOptions `json:"ipv6,omitempty"`
```
A similar suggestion to the previous one.

interfaces.go
```diff
@@ -0,0 +1,215 @@
...
+
+type InterfacesPublic struct {
+	IPv4 *InterfacesIPv4 `json:"ipv4,omitempty"`
+	IPv6 *InterfacesIPv6 `json:"ipv6,omitempty"`
+}
+
+type InterfacesIPv4 struct {
+	Addresses []InterfacesIPv4Address `json:"addresses"`
+	Shared    []InterfacesIPv4Shared  `json:"shared"`
+}
```
ChihweiLHBird, line 36: The POST object of ipv4 of a public interface doesn't have shared key..

zliang-akamai (approved):

ezilber-akamai (approved):
Nice work!

PR #725, opened 2025-04-18
Skip monthly transfer v1 test on 32 bits systems

zliang-akamai:
## 📝 Description

This is to ensure unit tests can go through on 32 bits system.


ezilber-akamai (approved):
Thanks for fixing this!

jriddle-linode (approved):
LGTM

dswarbrick:
This doesn't really do what was intended. The issue isn't that the test fails at _runtime_ on 32-bit.

The issue is that the code will fail to _build_ on 32-bit, due to the integer overflows.
```
# github.com/linode/linodego/test/unit [github.com/linode/linodego/test/unit.test]
test/unit/instance_test.go:132:18: cannot use 30471077120 (untyped int constant) as int value in argument to assert.Equal (overflows)
test/unit/instance_test.go:133:18: cannot use 22956600198 (untyped int constant) as int value in argument to assert.Equal (overflows)
test/unit/instance_test.go:134:18: cannot use 53427677318 (untyped int constant) as int value in argument to assert.Equal (overflows)
test/unit/mysql_test.go:404:18: cannot use 4294967295 (untyped int constant) as int value in argument to assert.Equal (overflows)
test/unit/mysql_test.go:412:18: cannot use 1099511627776 (untyped int constant) as int value in argument to assert.Equal (overflows)
FAIL    github.com/linode/linodego/test/unit [build failed]
```

What you need to do is separate this test into a file that will not be picked up by the compiler on 32-bit platforms, e.g. with a `//go:build !(386 || arm)` build tag.

(and yes, the failing `mysql_test.go` tests are new, but fail for the same reason).

PR #727, opened 2025-04-21
Added support for `networking/` and `vpcs/` endpoint changs for VPC Linodes Enhanced Interfaces

ezilber-akamai:
## 📝 Description

Added changes for VPC Linodes Enhanced Interfaces for `networking/` and `vpcs/` endpoints. Also added corresponding unit tests.

## ✔️ How to Test

`make test-unit`


firewalls.go
```diff
@@ -50,6 +51,31 @@ type FirewallUpdateOptions struct {
 	Tags   *[]string      `json:"tags,omitempty"`
 }
 
+// DefaultFirewalls represents the default firewalls for Linodes,
+// Linode VPC and public interfaces, and NodeBalancers.
+type DefaultFirewalls struct {
+	DefaultFirewallIDs DefaultFirewallIDs `json:"default_firewall_ids"`
+}
```
zliang-akamai, line 58: Should we rename this struct (and its fellow functions) to be `FirewallSettings` because there might be more settings in addition to `default_firewall_ids` to come?

ezilber-akamai: Done!

lgarber-akamai (approved):
Looks great and matches the API specification! This feature is _technically_ available in LA but I think it's fine for these integration tests to be implemented in a separate PR.

PR #736, opened 2025-04-30
fix: update MigratingTo field to accept int or null

cxdy:
## 📝 Description

The Linode API returns either an Integer or `null` for this field. We need to take that instead of a string.

```
level=error msg="An error occurred in redacted worker." func="main.Worker()" file="lipservice.go:45" err="[002] json: cannot unmarshal number into Go struct field paginatedResponse[github.com/linode/linodego.Instance].data.Mask.placement_group.migrating_to of type string"
```

## ✔️ How to Test

I don't know if I can say on Github, but hit me up on Slack if you want further detail here.

I didn't see any tests for PlacementGroups in the instances_test.go file, but I can probably add them if you'd like.


cxdy:
thanks for fixing the tests @zliang-akamai 🙂

ykim-akamai (approved):
LGTM, thanks for the fix!!

PR #751, opened 2025-05-15
project: UDP NodeBalancers

ezilber-akamai:
## 📝 Description

This pull request merges the NodeBalancer UDP project into this project's main branch to be included in the next release.

## ✔️ How to Test

The following test steps assume you have pulled down this PR locally and are using an account with access to UDP NodeBalancers.

### Unit Tests
`make test-unit`

### Integration Tests
`make fixtures TEST_ARGS="-run TestNodeBalancer"`

### DIFF
--- a/go.work.sum
+++ b/go.work.sum
@@ -57,7 +57,6 @@ golang.org/x/crypto v0.35.0/go.mod h1:dy7dXNW32cAb/6/PRuTNsix8T+vJAqvuIy5Bli/x0Y
 golang.org/x/crypto v0.36.0 h1:AnAEvhDddvBdpY+uR+MyHmuZzzNqXSe/GvuDeob5L34=
 golang.org/x/crypto v0.36.0/go.mod h1:Y4J0ReaxCR1IMaabaSMugxJES1EpwhBHhv2bDHklZvc=
 golang.org/x/crypto v0.37.0/go.mod h1:vg+k43peMZ0pUMhYmVAWysMK35e6ioLh3wB8ZCAfbVc=
-golang.org/x/crypto v0.38.0 h1:jt+WWG8IZlBnVbomuhg2Mdq0+BBQaHbtqHEFEigjUV8=
 golang.org/x/crypto v0.38.0/go.mod h1:MvrbAqul58NNYPKnOra203SB9vpuZW0e+RRZV+Ggqjw=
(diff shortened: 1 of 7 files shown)


jriddle-linode (approved):
LGTM tests are passing

nodebalancer_configs.go
```diff
@@ -160,6 +168,7 @@ func (i NodeBalancerConfig) GetCreateOptions() NodeBalancerConfigCreateOptions {
 		CheckPath:     i.CheckPath,
 		CheckBody:     i.CheckBody,
 		CheckPassive:  copyBool(&i.CheckPassive),
+		UDPCheckPort:  &i.UDPCheckPort,
```
rahulait, line 171: We should set it only if protocol is UDP. This field is not optional in `NodeBalancerConfig`, so if you don't set it there (ex for TCP protocol), it gets default 0. If then someone does `GetCreateOptions()` for that `NodeBalancerConfig`, returned `NodeBalancerConfigCreateOptions` has UDPCheckPort assigned value 0 and that will make calls to fail as udp_check_port has to be within 1 to 65535.

nodebalancer_configs.go
```diff
@@ -181,6 +190,7 @@ func (i NodeBalancerConfig) GetUpdateOptions() NodeBalancerConfigUpdateOptions {
 		CheckBody:     i.CheckBody,
 		CheckPassive:  copyBool(&i.CheckPassive),
 		CheckTimeout:  i.CheckTimeout,
+		UDPCheckPort:  copyInt(&i.UDPCheckPort),
```
rahulait, line 193: Same here, set only for UDP protocol

nodebalancer_configs.go
```diff
@@ -202,6 +212,7 @@ func (i NodeBalancerConfig) GetRebuildOptions() NodeBalancerConfigRebuildOptions
 		CheckPath:     i.CheckPath,
 		CheckBody:     i.CheckBody,
 		CheckPassive:  copyBool(&i.CheckPassive),
+		UDPCheckPort:  copyInt(&i.UDPCheckPort),
```
rahulait, line 215: Same here, set only for UDP protocol

nodebalancer_configs.go
ezilber-akamai: Thank you for the feedback! I added a check for this condition in `GetCreateOptions()`, so a 0 will no longer get passed to the API in this case.

lgarber-akamai (approved):
🎉

PR #753, opened 2025-05-20
Deprecate MarkEventRead API

zliang-akamai:
## 📝 Description

The `/account/events/{eventId}/read` endpoint will be deprecated on May 20th. Add a deprecation warning here.


lgarber-akamai (approved):
Confirmed that this endpoint will be deprecated alongside the API release today. Thanks for handling this!

ykim-akamai (approved):
Thank you for updating this!

PR #758, opened 2025-05-26
UseURL should require host and scheme

kokes:
## 📝 Description

Calling `UseURL` with `foo.bar.com` results in the client being configured with `api.linode.com` silently, because `foo.bar.com` (without a scheme) gets parsed into a URL object, where `foo.bar.com` **is the path** (and [everything else is empty](https://go.dev/play/p/olSNQ3JLVuO)), which is then **ignored** by our implementation: https://github.com/linode/linodego/blob/0bb8fc2ae35f2c4a60ff7a549db141335d30c1af/client.go#L535-L539

Since we're leveraging just the Host and Scheme portions of the URL object, I introduced a check for these values. Also added tests to cover various scenarios.

(Note that I not only converted existing UseURL tests to a single table test, but I also changed the logic where I now test against `c.resty.BaseURL`, which is the URL actually used for making requests - it originally tested against `c.baseURL`, which can be misleading.)

## ✔️ How to Test

Run the following before checking out the changes from this PR:

```go
package main

import (
	"context"
	"log"

	"github.com/linode/linodego"
)

func main() {

	// This is intentionally set to an invalid URL
	if _, err := client.UseURL("api.test1.linode.com/v4beta"); err != nil {
		log.Fatal(err)
	}

	if _, err := client.ListInstances(context.Background(), nil); err != nil {
		log.Fatal(err)
	}
}
```

You'd be greeted with a `[401] Invalid Token`, even though you think you're accessing a non-existent endpoint. It's actually contacting production.

Check out the PR and try again, you'll be met with `need both scheme and host in API URL, got "api.test1.linode.com/v4beta"`

As usual, I just added a few unit tests.


jriddle-linode (approved):
Great update thank you!

PR #760, opened 2025-05-28
Firewall Related integration tests 

vshanthe:
## 📝 Description
Firewall setting related Tests

## ✔️ How to Test
make fixtures TEST_ARGS="-run TestFirewallSettings"

## 📷 Preview

### DIFF
--- a/test/integration/firewalls_test.go
+++ b/test/integration/firewalls_test.go
@@ -273,7 +273,6 @@ func TestFirewallSettings_UpdateAllFields(t *testing.T) {
 	}
 }
 
-
 func TestFirewallSettings_UpdatePartial(t *testing.T) {
 	label := fmt.Sprintf("fw-partial-%s", getUniqueText())
 


test/integration/firewalls_test.go
```diff
@@ -164,3 +165,119 @@ func setupFirewall(t *testing.T, firewallModifiers []firewallModifier, fixturesY
...
+		settings.DefaultFirewallIDs.VPCInterface == 0 {
+		t.Log("No default firewall IDs set — this is acceptable in a fresh test environment.")
+	}
+}
+
+func TestFirewallSettings_UpdateAllFields(t *testing.T) {
+	label := fmt.Sprintf("fw-allfields-%s", getUniqueText())
+	if len(label) < 3 || len(label) > 32 {
+		t.Fatalf("generated label %q is %d chars; must be 3–32", label, len(label))
+	}
```
zliang-akamai:
```suggestion
```
`getUniqueText` always return a string with length of 19, so I think we can safely remove this check.

test/integration/firewalls_test.go
```diff
@@ -164,3 +165,119 @@ func setupFirewall(t *testing.T, firewallModifiers []firewallModifier, fixturesY
...
+
+	client, created, teardown, err := setupFirewall(t, []firewallModifier{
```
zliang-akamai:
```suggestion
	client, firewall, teardown, err := setupFirewall(t, []firewallModifier{
```
A suggestion for a more straight forward naming of the var

test/integration/firewalls_test.go
```diff
@@ -164,3 +165,119 @@ func setupFirewall(t *testing.T, firewallModifiers []firewallModifier, fixturesY
...
+	if err != nil {
+		t.Fatalf("Error getting firewall settings: %v", err)
+	}
+
+	if settings.DefaultFirewallIDs.Linode == 0 &&
+		settings.DefaultFirewallIDs.NodeBalancer == 0 &&
+		settings.DefaultFirewallIDs.PublicInterface == 0 &&
+		settings.DefaultFirewallIDs.VPCInterface == 0 {
+		t.Log("No default firewall IDs set — this is acceptable in a fresh test environment.")
+	}
```
ykim-akamai, line 210: Should there be more assertions on all the values in `settings` are set correctly on this get request?

test/integration/firewalls_test.go
```diff
@@ -164,3 +165,119 @@ func setupFirewall(t *testing.T, firewallModifiers []firewallModifier, fixturesY
...
+			opts.Label = label
+		},
+	}, "fixtures/TestFirewallSettings_UpdateAllFields")
+	if err != nil {
+		t.Fatal(err)
+	}
+	if created == nil {
+		teardown()
+		t.Fatal("setupFirewall returned nil firewall")
+	}
```
zliang-akamai:
```suggestion
```

We usually assume when there is not an error, `setupFirewall` always return a valid firewall. So I think we can cleanup this check as well. (Even in the extreme case that the assumption is wrong, it will still fail by a nil pointer panic in the lines below)

test/integration/firewalls_test.go
```diff
@@ -164,3 +165,119 @@ func setupFirewall(t *testing.T, firewallModifiers []firewallModifier, fixturesY
...
+
+	defer func() {
+		updateOpts := linodego.FirewallSettingsUpdateOptions{
+			DefaultFirewallIDs: linodego.DefaultFirewallIDsOptions{
```
ykim-akamai, line 234: We should update with some values here and assert on if it is reflected in the response

test/integration/firewalls_test.go
```diff
@@ -164,3 +165,119 @@ func setupFirewall(t *testing.T, firewallModifiers []firewallModifier, fixturesY
...
+				NodeBalancer:    nil,
+				PublicInterface: nil,
+				VPCInterface:    nil,
+			},
+		}
+		_, err := client.UpdateFirewallSettings(context.Background(), updateOpts)
+		if err != nil {
+			t.Fatalf("failed to unset default firewall IDs: %v", err)
+		}
+	}()
```
zliang-akamai: We won't be able to unset the default firewall to null. Thus, setting it back to original default firewall of the account prior to the test may be better. And in the case that the account wasn't with a default firewall setting, I think we can leave it as is.

ykim-akamai (approved):

PR #763, opened 2025-06-04
Added support for VPC DBaaS Integration

ezilber-akamai:
## 📝 Description

Added support for VPC/DBaaS Integration features and unit tests.

## ✔️ How to Test

### Unit Testing
`make test-unit`

### DIFF
--- a/postgres.go
+++ b/postgres.go
@@ -69,6 +69,7 @@ type PostgresDatabase struct {
 	UsedDiskSizeGB    int                          `json:"used_disk_size_gb"`
 	TotalDiskSizeGB   int                          `json:"total_disk_size_gb"`
 	EngineConfig      PostgresDatabaseEngineConfig `json:"engine_config"`
+	PrivateNetwork    *DatabasePrivateNetwork      `json:"private_network,omitempty"`
 }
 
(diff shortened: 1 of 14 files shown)


postgres.go
```diff
@@ -612,18 +613,20 @@ type PostgresCreateOptions struct {
 
 	Fork *DatabaseFork `json:"fork,omitempty"`
 
-	EngineConfig *PostgresDatabaseEngineConfig `json:"engine_config,omitempty"`
+	EngineConfig   *PostgresDatabaseEngineConfig `json:"engine_config,omitempty"`
+	PrivateNetwork *DatabasePrivateNetwork       `json:"private_network,omitempty"`
```
lgarber-akamai, line 617: Normally I'd be a concerned about whether this type of substruct would diverge per-database type in the future, but I don't think that's likely in this case.

lgarber-akamai (approved):
Great work!

zliang-akamai (approved):
LGTM!

postgres.go
ezilber-akamai: Yeah I thought separating it out into two separate structs, but I agree its not likely to diverge.

PR #764, opened 2025-06-05
Add `DoublePointer` helper function

zliang-akamai:
## 📝 Description

Add `DoublePointer` helper function for some APIs that treats omitted attribute and null value attribute differently.

## ✔️ How to Test

```bash
make test-unit
```

PR #765, opened 2025-06-05
Make default firewall IDs pointers

zliang-akamai:
## 📝 Description

Default firewall ID returning values are nullable so making them as pointers.

The API will treat a `null` value and excluding the attribute in the request body differently, so making them double pointers.

## ✔️ How to Test
```bash
make test-unit
```


zliang-akamai:
A part of this PR is separated to https://github.com/linode/linodego/pull/764

ezilber-akamai (approved):
LGTM!

lgarber-akamai (approved):

PR #768, opened 2025-06-12
feat: add Marshall for token

guilhem:
## 📝 Description

Add Marshal JSON function to token.
This primarily helps when testing an application to match API return.

## ✔️ How to Test

## 📷 Preview


zliang-akamai:
Will it be possible to implement it outside of the linodego code base because most structs in linodego don't have a customized Marshall function?

PR #771, opened 2025-06-18
Addressed additional API Spec changes

ezilber-akamai:
## 📝 Description

Addressed recent API Spec changes and updated unit tests

## ✔️ How to Test

`make test-unit`


lgarber-akamai (approved):
Looks great and seems to align with the API changes doc. It is a bit weird that the event action is `linode_poweroff_on` while the maintenance policy string is `linode/power_off_on`, but it's documented that way internally so this PR should be good to go!

PR #773, opened 2025-06-24
Make interface optional fields pointers

zliang-akamai:
## 📝 Description

Some optional fields were previously and incorrectly added as non-pointer fields. Correct them in this PR.

This PR is based on https://github.com/linode/linodego/pull/772 so there are some irrelevant test fixes from the changes from that branch.

## ✔️ How to Test
`make test-int`


yec-akamai (approved):
LGTM!

PR #775, opened 2025-07-01
Add future breaking change notice for ConfigID in VPC IP

zliang-akamai:
## 📝 Description

ConfigID is now nullable from the API, and needs to be changed to a pointer in the next major version. So adding a notice here.

The lint error reported by the CI is irrelevant, and should be fixed from the main branch.

### DIFF
--- a/instance_ips.go
+++ b/instance_ips.go
@@ -48,8 +48,10 @@ type VPCIP struct {
 	NAT1To1      *string `json:"nat_1_1"`
 	VPCID        int     `json:"vpc_id"`
 	SubnetID     int     `json:"subnet_id"`
-	ConfigID     int     `json:"config_id"`
 	InterfaceID  int     `json:"interface_id"`
+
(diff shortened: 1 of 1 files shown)


instance_ips.go
```diff
@@ -48,8 +48,10 @@ type VPCIP struct {
 	NAT1To1      *string `json:"nat_1_1"`
 	VPCID        int     `json:"vpc_id"`
 	SubnetID     int     `json:"subnet_id"`
-	ConfigID     int     `json:"config_id"`
 	InterfaceID  int     `json:"interface_id"`
+
+	// Scheduled breaking change: ConfigID is now nullable from the API and will become a pointer in the next major version.
```
lgarber-akamai: Just to make this a bit more understandable for external users, would it make sense to reword this to something like:

```
The type of this field will be made a pointer in the next major release of linodego.
```

Otherwise looks great!

lgarber-akamai (approved):
LGTM!

PR #776, opened 2025-07-02
Nullable default firewall ids

zliang-akamai:
## 📝 Description

Making `DefaultFirewallIDs` field a pointer, so we can omit it in the future if other firewall settings are added.

## ✔️ How to Test
`make test-int`


ezilber-akamai (approved):

ykim-akamai (approved):
Thanks for addressing this

PR #777, opened 2025-07-02
Support ACLP fetching entity metrics

yec-akamai:
## 📝 Description

This PR introduced a new client MonitorClient, to make calls to a different host monitor-api. It utilizes PAT token generated by linode api. The endpoint for this client is fetching metrics for a list of entities. Also, updated test suite for the new client and test the fetching metrics endpoint.

### Unit test:

```
make test-unit TEST_ARGS="-run TestFetchEntityMetrics" 
```

### Integration test:

```
make test-int TEST_ARGS="-run TestMonitorAPI_Fetch_Entity_Metrics"
```

### DIFF
--- a/test/integration/fixtures/TestInstance_GetMonthlyTransfer.yaml
+++ b/test/integration/fixtures/TestInstance_GetMonthlyTransfer.yaml
@@ -17,24 +17,27 @@ interactions:
     body: '{"data": [{"id": "ap-west", "label": "Mumbai, IN", "country": "in", "capabilities":
       ["Linodes", "LA Disk Encryption", "Disk Encryption", "Backups", "NodeBalancers",
       "Block Storage", "GPU Linodes", "Kubernetes", "Cloud Firewall", "Vlans", "Block
-      Storage Migrations", "Managed Databases", "Metadata", "Placement Group", "StackScripts",
-      "Linode Interfaces"], "status": "ok", "resolvers": {"ipv4": "172.105.34.5,172.105.35.5,172.105.36.5,172.105.37.5,172.105.38.5,172.105.39.5,172.105.40.5,172.105.41.5,172.105.42.5,172.105.43.5",
+      Storage Migrations", "Managed Databases", "Metadata", "Placement Group", "StackScripts"],
+      "monitors": {"alerts": ["Managed Databases"], "metrics": ["Managed Databases"]},
+      "status": "ok", "resolvers": {"ipv4": "172.105.34.5,172.105.35.5,172.105.36.5,172.105.37.5,172.105.38.5,172.105.39.5,172.105.40.5,172.105.41.5,172.105.42.5,172.105.43.5",
       "ipv6": "1234::5678,1234::5678,1234::5678,1234::5678,1234::5678,1234::5678,1234::5678,1234::5678,1234::5678,1234::5678"},
       "placement_group_limits": {"maximum_pgs_per_customer": null, "maximum_linodes_per_pg":
       5, "maximum_linodes_per_flexible_pg": 5}, "site_type": "core"}, {"id": "ca-central",
       "label": "Toronto, CA", "country": "ca", "capabilities": ["Linodes", "LA Disk
       Encryption", "Disk Encryption", "Backups", "NodeBalancers", "Block Storage",
       "Kubernetes", "Cloud Firewall", "Vlans", "Block Storage Migrations", "Managed
-      Databases", "Metadata", "Placement Group", "StackScripts", "Linode Interfaces"],
-      "status": "ok", "resolvers": {"ipv4": "172.105.0.5,172.105.3.5,172.105.4.5,172.105.5.5,172.105.6.5,172.105.7.5,172.105.8.5,172.105.9.5,172.105.10.5,172.105.11.5",
+      Databases", "Metadata", "Placement Group", "StackScripts"], "monitors": {"alerts":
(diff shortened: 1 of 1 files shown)


client_monitor.go
```diff
@@ -0,0 +1,169 @@
...
+	"net/url"
+	"os"
+	"path"
+
+	"github.com/go-resty/resty/v2"
+)
+
+const (
+	// MonitorAPIHost is the default monitor-api host
+	MonitorAPIHost = "monitor-api.linode.com"
```
zliang-akamai, line 16: Although it's different from existing naming, do you think it would be better to be `DefaultMonitorAPIHost` to reflect its real usage?

client_monitor.go
```diff
@@ -0,0 +1,169 @@
...
+	if mc.apiProtocol != "" {
+		apiProto = mc.apiProtocol
+	}
+
+	mc.resty.SetBaseURL(
+		fmt.Sprintf(
+			"%s://%s/%s",
+			apiProto,
+			baseURL,
+			url.PathEscape(apiVersion),
```
zliang-akamai, line 166: Some endpoints doesn't seem to have API version in their path in the spec. Can we verify adding the API version won't break them?

client_monitor.go
yec-akamai: I think it's better to keep them aligned for maintenance. Also the all the constant value are a default value, so it feels not necessary to add a default prefix here. In this case, we can keep the constant name short and clear.

client_monitor.go
yec-akamai: Which endpoints are you refering to? AFAIK there is only one endpoint currently for this client: https://techdocs.akamai.com/linode-api/reference/post-read-metric

zliang-akamai: Just realized all other APIs are on Linode APIv4, so this won't be a problem. Sorry about the confusion.

monitor_api_services.go
```diff
@@ -0,0 +1,121 @@
...
+}
+
+// MetricAbsoluteTimeDuration specifies an absolute time range for data queries
+type MetricAbsoluteTimeDuration struct {
+	Start time.Time `json:"start,omitempty"`
+	End   time.Time `json:"end,omitempty"`
+}
+
+// FetchEntityMetrics returns metrics information for the individual entities within a specific service type
+func (mc *MonitorClient) FetchEntityMetrics(ctx context.Context, serviceType string, opts *EntityMetricsFetchOptions) (*EntityMetrics, error) {
```
zliang-akamai, line 101:
```suggestion
// GetEntityMetrics returns metrics information for the individual entities within a specific service type
func (mc *MonitorClient) GetEntityMetrics(ctx context.Context, serviceType ServiceType, opts *GetEntityMetricsOptions) (*EntityMetrics, error) {
```

Should we change `Fetch` to `Get` to be consistent with the [API doc](https://techdocs.akamai.com/linode-api/reference/post-read-metric)?

And I think the type should be the alias of string `ServiceType` so users can use `linodego.ServiceTypeDBaaS` without casting it into string (`string(linodego.ServiceTypeDBaaS)`), and all other monitor functions may need this change as well.

test/unit/monitor_api_services_test.go
```diff
@@ -0,0 +1,56 @@
...
+	defer base.TearDown(t)
+
+	serviceType := "dbaas"
+	base.MockPost(fmt.Sprintf("monitor/services/%s/metrics", serviceType), fixtureData)
+
+	opts := linodego.EntityMetricsFetchOptions{
+		EntityIDs: []any{13217, 13316},
+		Metrics: []linodego.EntityMetric{
+			{
+				Name:              "avg_read_iops",
```
zliang-akamai: Couldn't get this metric name work with a real API call, but `read_iops` works. Similarly `cpu_usage` metric name works but `avg_cpu_usage` doesn't. Is there any way to make have a metric with name `avg_read_iops` in the real API call? Otherwise, can we replace these metric names in the unit test to better reflect the real API?

monitor_api_services.go
yec-akamai: Using `fetch` here is because this endpoint is actually a `POST`. However, the prefix `Get` usually indicates an actual `GET` endpoint, which could be confusing. I've talked to the ACLP team about this situation, and they suggested `fetch` might be a better way to go. 

I'm not sure if using a `ServiceType` string is better here. Because the `ServiceType` enum includes a list of services, however, here we only accept dbaas for now. It could be misleading that we also support other service types.

test/unit/monitor_api_services_test.go
yec-akamai: I will replace the name here

monitor_api_services.go
zliang-akamai: Okay, `Fetch` seems to be fine then. Maybe there can be a chance to update the doc though. But `ServiceType` is defined for this usage, right? Meaning we will get those services supported in the future?
Maybe checking with ACLP team for their opinion?

zliang-akamai (approved):
Looks good and works well!

monitor_api_services.go
yec-akamai: The `ServiceType` was defined for another set of endpoints, Dashboard. It's a separate effort though. From what I've learnt there isn't a clear plan for what services will be support in the next for Metrics here.

jriddle-linode (approved):
Looks good on my end and is working locally.

PR #781, opened 2025-07-08
Add integration test for VM Maintenance Policy changes

vshanthe:
## 📝 Description
 integration tests
Test Against Devcloud using API TOKEN
and URL - api.devcloud.linode.com

## ✔️ How to Test
make fixtures TEST_ARGS="-run TestMaintenancePolicies_List"
make fixtures TEST_ARGS="-run TestAccountSettings"
make fixtures TEST_ARGS="-run TestInstance_MaintenancePolicy"

Note:  Integration Test will fail since changes are available in Devcloud 
## 📷 Preview


test/integration/instances_test.go
```diff
@@ -1019,3 +1019,59 @@ func setupInstanceWithoutDisks(t *testing.T, fixturesYaml string, enableCloudFir
 	}
 	return client, instance, config, teardown, err
 }
+
+func TestInstance_MaintenancePolicy(t *testing.T) {
+	client, teardown := createTestClient(t, "fixtures/TestInstance_MaintenancePolicy")
+	defer teardown()
+
+	// test default maintenance policy (should be empty string)
```
ezilber-akamai: The default maintenance policy for instances should be the default maintenance policy for the account, so maybe we can call `GET /account/settings` and compare with that? Comparing to `linode/migrate` could cause intermittent failures if the policy ever changes on our test accounts.

test/integration/account_maintenance_test.go
```diff
@@ -17,3 +17,32 @@ func TestAccountMaintenances_List(t *testing.T) {
 		t.Errorf("Error listing maintenances, expected array, got error %v", err)
 	}
 }
+
+func TestAccountMaintenancePolicies_List(t *testing.T) {
```
ezilber-akamai: This tests should probably be in its own file (`maintenance_policies_test.go` or something like that) since the endpoint it is testing does not fall under the `/account` resource. The name can also be changed to `TestMaintenancePolicies_List` for clarity.

ezilber-akamai (commented):
Looks good! There are a few more integration tests listed in the test plan that I think would be worth adding:
- We should add checks for `GET /linode/instances`, `GET /linode/instances/{linodeId}`, and `POST /linode/instances/{linodeId}/clone`

Other than that, there are a couple other nitpicks I added comments for, but looks great overall!

test/integration/maintenance_test.go
```diff
@@ -0,0 +1,50 @@
...
+import (
+	"context"
+	"encoding/json"
+	"testing"
+
+	"github.com/linode/linodego"
+	"github.com/stretchr/testify/require"
+)
+
+func TestAccountMaintenancePolicies_List(t *testing.T) {
```
ezilber-akamai: Can you rename this test and its fixture from `TestAccountMaintenancePolicies_List` to `TestMaintenancePolicies_List`?

ezilber-akamai: Also one other thing I just noticed. I think we should be updating `TestInstance_Clone` rather than `TestInstance_Disk_Clone`. `TestInstance_Clone` is one of the endpoints impacted by this project, but `TestInstance_Disk_Clone` is not.

test/integration/maintenance_test.go
```diff
@@ -0,0 +1,50 @@
...
+	"context"
+	"encoding/json"
+	"testing"
+
+	"github.com/linode/linodego"
+	"github.com/stretchr/testify/require"
+)
+
+func TestMaintenancePolicies_List(t *testing.T) {
+	client, fixtureTeardown := createTestClient(t, "fixtures/TestAccountMaintenancePolicies_List")
```
ezilber-akamai:
```suggestion
	client, fixtureTeardown := createTestClient(t, "fixtures/TestMaintenancePolicies_List")
```

We should change the name of the fixture to align with the new name of the test (and delete the old fixture)

Other than that everything looks good!

zliang-akamai (approved):

ezilber-akamai (approved):

PR #786, opened 2025-07-21
Add label field to lkenodepool

sphan-akamai:
## 📝 Description

This PR adds the `Label` field to lkenodepool. LKE recently added support for lkenodepool naming via the `Label` field. (ex., when a customer provides a `Label` during the creation/update of a lkenodepool, nodes in that nodepool will following this naming convention `<lkenodepool namespace>-<lkenodepool label>-<5 char hash>`).

## ✔️ How to Test

Perform the following steps for both Enterprise Clusters and Non-Enterprise Clusters

Validate that creating a nodepool with Label set, results in the nodepool following this naming convention: `<lkenodepool namespace>-<lkenodepool label>-<5 char hash>`)
1. Create a cluster 
2. Create a nodepool for the cluster you created in step 1 with `Label` set in the POST request
3. Verify that the nodepool created in step 2 follows this naming convention: `<lkenodepool namespace>-<lkenodepool label>-<5 char hash>`) where `<lkenodepool label>` is equal to whatever value you provided to `Label` in step 2

Validate that un-setting the `Label` field, results in the nodepool following the original naming convention: `{nodepoolnamespace}-{nodepoolID}-{5 char hash}-{5 char hash}`
1. Using the nodepool you created above, update the nodepool via PUT request with `Label` set to an empty string 
2. Validate that it is named correctly (you can increase the replica count of the nodepool to check that new nodes in the nodepool use the correct naming convention*)

***Note**: Customers who modify existing nodepools will not see changes to the nodepool names after setting the Label until they recycle their nodepools

### Unit Tests 

Run `make test-unit TEST_ARGS='-run TestLKENodePool'`
Run `make test-unit TEST_ARGS='-run TestLKEEnterpriseNodePool`

### DIFF
--- a/lke_node_pools.go
+++ b/lke_node_pools.go
@@ -93,7 +93,7 @@ type LKENodePoolCreateOptions struct {
 	Tags   []string           `json:"tags"`
 	Labels LKENodePoolLabels  `json:"labels"`
 	Taints []LKENodePoolTaint `json:"taints"`
-	Label  *string            `json:"label"`
+	Label  *string            `json:"label,omitempty"`
 
 	Autoscaler *LKENodePoolAutoscaler `json:"autoscaler,omitempty"`
 
@@ -109,7 +109,7 @@ type LKENodePoolUpdateOptions struct {
 	Tags   *[]string           `json:"tags,omitempty"`
 	Labels *LKENodePoolLabels  `json:"labels,omitempty"`
 	Taints *[]LKENodePoolTaint `json:"taints,omitempty"`
-	Label  *string             `json:"label"`
+	Label  *string             `json:"label,omitempty"`
 
 	Autoscaler *LKENodePoolAutoscaler `json:"autoscaler,omitempty"`
 


lke_node_pools.go
```diff
@@ -92,6 +93,7 @@ type LKENodePoolCreateOptions struct {
 	Tags   []string           `json:"tags"`
 	Labels LKENodePoolLabels  `json:"labels"`
 	Taints []LKENodePoolTaint `json:"taints"`
+	Label  string             `json:"label,omitempty"`
```
zliang-akamai: Can we make it as a pointer to distinguish "" and field omitted in actual API call?

lke_node_pools.go
```diff
@@ -72,6 +72,7 @@ type LKENodePool struct {
 	Tags    []string            `json:"tags"`
 	Labels  LKENodePoolLabels   `json:"labels"`
 	Taints  []LKENodePoolTaint  `json:"taints"`
+	Label   *string             `json:"label,omitempty"`
```
zliang-akamai:
```suggestion
	Label   *string             `json:"label"`
```
Can you remove the `omitempty` tag here? This struct only represents the response and won't be used in an API request so it's not needed.

lke_node_pools.go
```diff
@@ -92,6 +93,7 @@ type LKENodePoolCreateOptions struct {
 	Tags   []string           `json:"tags"`
 	Labels LKENodePoolLabels  `json:"labels"`
 	Taints []LKENodePoolTaint `json:"taints"`
+	Label  *string            `json:"label"`
```
jriddle-linode: Sorry for any confusion but Create and Update do need the `omitempty` tag if the field is nullable. Only the `LKENodePool` struct doesn't need it because its a response and not a request.

lke_node_pools.go
```diff
@@ -107,6 +109,7 @@ type LKENodePoolUpdateOptions struct {
 	Tags   *[]string           `json:"tags,omitempty"`
 	Labels *LKENodePoolLabels  `json:"labels,omitempty"`
 	Taints *[]LKENodePoolTaint `json:"taints,omitempty"`
+	Label  *string             `json:"label"`
```
jriddle-linode: ++

sphan-akamai: Understood! fixing now.

PR #800, opened 2025-08-18
Added support for Private Image Sharing

ezilber-akamai:
## 📝 Description

Added support for new Private Image Sharing features and corresponding unit tests.

## ✔️ How to Test

`make test-unit`


image_sharegroups_consumer.go
```diff
@@ -0,0 +1,208 @@
...
+type ImageShareGroupToken struct {
+	TokenUUID              string     `json:"token_uuid"`
+	Status                 string     `json:"status"`
+	Label                  string     `json:"label"`
+	ValidForShareGroupUUID string     `json:"valid_for_sharegroup_uuid"`
+	Created                *time.Time `json:"-"`
+	Updated                *time.Time `json:"-"`
+	Expiry                 *time.Time `json:"-"`
+	ShareGroupUUID         *string    `json:"sharegroup_uuid,omitempty"`
+	ShareGroupLabel        *string    `json:"sharegroup_label,omitempty"`
```
zliang-akamai: Is `omitempty` unnecessary on these two fields as this struct won't be used to build a request body?

images.go
```diff
@@ -43,16 +43,17 @@ type ImageRegion struct {
...
 type Image struct {
 	ID           string        `json:"id"`
-	CreatedBy    string        `json:"created_by"`
+	CreatedBy    *string       `json:"created_by,omitempty"`
 	Capabilities []string      `json:"capabilities"`
 	Label        string        `json:"label"`
 	Description  string        `json:"description"`
 	Type         string        `json:"type"`
-	Vendor       string        `json:"vendor"`
+	Vendor       *string       `json:"vendor,omitempty"`
```
zliang-akamai: I think we may also remove `omitempty` here since they won't be used to build a request? By the way, just curious, are they becoming nullable so we want to make them as pointers?

ezilber-akamai: Good point about them not needing to use `omitempty`, not sure what I was thinking when I wrote this 😆. But yes, these fields are becoming nullable so we should make them pointers.

image_sharegroups_consumer.go
ezilber-akamai: Good point. I will fix this.

zliang-akamai (approved):

images.go
```diff
@@ -61,6 +62,25 @@ type Image struct {
 	Created *time.Time `json:"-"`
 	Expiry  *time.Time `json:"-"`
 	EOL     *time.Time `json:"-"`
+
+	ImageSharing ImageSharing `json:"image_sharing"`
+}
+
+type ImageSharing struct {
+	SharedWith *ImageSharingSharedWith `json:"shared_with,omitempty"`
+	SharedBy   *ImageSharingSharedBy   `json:"shared_by,omitempty"`
```
yec-akamai: These two fields seem also for response struct only, we probably can remove `omitempty` as well

images.go
```diff
@@ -61,6 +62,25 @@ type Image struct {
...
+type ImageSharingSharedWith struct {
+	ShareGroupCount        int    `json:"sharegroup_count"`
+	ImageShareGroupListURL string `json:"image_sharegroup_list_url"`
+}
+
+type ImageSharingSharedBy struct {
+	ShareGroupID    int     `json:"sharegroup_id"`
+	ShareGroupUUID  string  `json:"sharegroup_uuid"`
+	ShareGroupLabel string  `json:"sharegroup_label"`
+	SourceImageID   *string `json:"source_image_id,omitempty"`
```
yec-akamai: Same for this SourceImageID field

images.go
```diff
@@ -43,16 +43,17 @@ type ImageRegion struct {
...
 	Label        string        `json:"label"`
 	Description  string        `json:"description"`
 	Type         string        `json:"type"`
-	Vendor       string        `json:"vendor"`
+	Vendor       *string       `json:"vendor"`
 	Status       ImageStatus   `json:"status"`
 	Size         int           `json:"size"`
 	TotalSize    int           `json:"total_size"`
 	IsPublic     bool          `json:"is_public"`
+	IsShared     *bool         `json:"is_shared"`
```
yec-akamai, line 56: Just a side note, I wonder how API would rather return nil instead of false for a bool value 🤔

image_sharegroups_consumer.go
```diff
@@ -0,0 +1,208 @@
...
+	t.Expiry = (*time.Time)(p.Expiry)
+
+	return nil
+}
+
+// ImageShareGroupCreateTokenResponse represents the response when the consumer
+// creates a single-use ImageShareGroup membership token.
+// The token itself is only provided upon creation, and must be given to the producer
+// via an outside medium for the consumer to be added as a member of the producer's ImageShareGroup.
+type ImageShareGroupCreateTokenResponse struct {
```
yec-akamai, line 88: This type seems very similar to the `ImageShareGroupToken` aside from an extra `Token` field. Do you think we can have it embedded like: 
```
type ImageShareGroupCreateTokenResponse struct {
  Token                  string     `json:"token"`
  ImageShareGroupToken
}
```

yec-akamai (commented):
Nice work! Just a few small things:

image_sharegroups_consumer.go
ezilber-akamai: I considered doing this but I decided to have them be separate structs in the unlikely event that they deviate in the future.

images.go
ezilber-akamai: According to a discussion in slack, the only situation in which `is_shared` is not nil will be when looking at a private image that has been shared.

images.go
ezilber-akamai: Good catch! Fixed

images.go
ezilber-akamai: Fixed!

images.go
yec-akamai: Got it!

yec-akamai (approved):
Unit tests passed locally!

PR #807, opened 2025-08-25
aclp region and group_by changes

pmajali:
## 📝 Description

1. Adding Monitor in the regions endpoint. This specifies the monitor services enabled in a particular region.
2. Adding group_by and filters field to the dashboard widget.
3. Enhancing services endpoint to include alert

## ✔️ How to Test

```
package main

import (
	"context"
	"fmt"
	"log"
	"net/http"
	"os"

	"github.com/linode/linodego"

	"golang.org/x/oauth2"
)

func main() {
	if !ok {
	}

	oauth2Client := &http.Client{
		Transport: &oauth2.Transport{
			Source: tokenSource,
		},
	}

	linodeClient.SetDebug(true)

	res, err := linodeClient.ListMonitorServices(context.Background(), nil)
	if err != nil {
		log.Fatal(err)
	}
	fmt.Printf("%v", res)

	res, err := linodeClient.ListRegions(context.Background(), nil)
	if err != nil {
		log.Fatal(err)
	}
	fmt.Printf("%v", res)
}
```

**unit tests**
make test-unit TEST_ARGS="-run TestListMonitorServices" 
make test-unit TEST_ARGS="-run TestListMonitorDashboards"
make test-unit TEST_ARGS="-run TestListRegions"

**Integration test**
make test-int TEST_ARGS="-run TestMonitorServices_Get_smoke"
make test-int TEST_ARGS="-run TestMonitorDashboards_Get_smoke"
make test-int TEST_ARGS="-run TestRegionsMonitorsSection"

### DIFF
--- a/monitor_dashboards.go
+++ b/monitor_dashboards.go
@@ -31,6 +31,7 @@ const (
 	ServiceTypeObjectStorage   ServiceType = "objectstorage"
 	ServiceTypeVPC             ServiceType = "vpc"
 	ServiceTypeFirewallService ServiceType = "firewall"
+	ServiceTypeNetLoadBalancer ServiceType = "netloadbalancer"
 )
 
 // enum object for DashboardType
@@ -51,6 +52,15 @@ type DashboardWidget struct {
 	ChartType         ChartType         `json:"chart_type"`
 	YLabel            string            `json:"y_label"`
 	AggregateFunction AggregateFunction `json:"aggregate_function"`
+	GroupBy           []string          `json:"group_by"`
+	Filters           []DashboardFilter `json:"filters,omitempty"`
+}
+
+// DashboardFilter represents a filter for dashboard widgets
+type DashboardFilter struct {
+	DimensionLabel string `json:"dimension_label"`
+	Operator       string `json:"operator"`
+	Value          string `json:"value"`
 }
 
 // Enum object for AggregateFunction
--- a/monitor_metrics_definitions.go
+++ b/monitor_metrics_definitions.go
@@ -45,6 +45,11 @@ const (
 	MetricUnitRatio          MetricUnit = "ratio"
 	MetricUnitOpsPerSecond   MetricUnit = "ops_per_second"
 	MetricUnitIops           MetricUnit = "iops"
+	MetricUnitKiloBytesPerSecond MetricUnit = "kilo_bytes_per_second"
+	MetricUnitSessionsPerSecond MetricUnit = "sessions_per_second"
+	MetricUnitPacketsPerSecond MetricUnit = "packets_per_second"
+	MetricUnitKiloBitsPerSecond MetricUnit = "kilo_bits_per_second"
+
 )
 
 // MonitorDimension represents an ACLP MonitorDimension object
--- a/monitor_services.go
+++ b/monitor_services.go
@@ -6,8 +6,16 @@ import (
 
 // MonitorService represents a MonitorService object
 type MonitorService struct {
-	Label       string `json:"label"`
-	ServiceType string `json:"service_type"`
+	Label       string               `json:"label"`
+	ServiceType string               `json:"service_type"`
+	Alert       *MonitorServiceAlert `json:"alert"`
+}
+
+// MonitorServiceAlert represents the alert configuration for a monitor service
+type MonitorServiceAlert struct {
+	PollingIntervalSeconds   []int    `json:"polling_interval_seconds"`
+	EvaluationPeriodSeconds  []int    `json:"evaluation_period_seconds"`
+	Scope                    []string `json:"scope"`
 }
 
 // ListMonitorServices lists all the registered ACLP MonitorServices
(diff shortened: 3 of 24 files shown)


monitor_dashboards.go
```diff
@@ -51,6 +52,15 @@ type DashboardWidget struct {
 	ChartType         ChartType         `json:"chart_type"`
 	YLabel            string            `json:"y_label"`
 	AggregateFunction AggregateFunction `json:"aggregate_function"`
+	GroupBy           []string          `json:"group_by"`
+	Filters           []DashboardFilter `json:"filters,omitempty"`
```
zliang-akamai: Response struct doesn't require `omitempty` tag

monitor_metrics_definitions.go
```diff
@@ -45,6 +45,11 @@ const (
 	MetricUnitRatio          MetricUnit = "ratio"
 	MetricUnitOpsPerSecond   MetricUnit = "ops_per_second"
 	MetricUnitIops           MetricUnit = "iops"
+	MetricUnitKiloBytesPerSecond MetricUnit = "kilo_bytes_per_second"
+	MetricUnitSessionsPerSecond MetricUnit = "sessions_per_second"
+	MetricUnitPacketsPerSecond MetricUnit = "packets_per_second"
+	MetricUnitKiloBitsPerSecond MetricUnit = "kilo_bits_per_second"
```
zliang-akamai: Can you format them to align with other constants above?

monitor_dashboards.go
pmajali: updated the struct

monitor_metrics_definitions.go
pmajali: updated to align with other units

monitor_services.go
```diff
@@ -6,17 +6,32 @@ import (
...
 
 // ListMonitorServices lists all the registered ACLP MonitorServices
 func (c *Client) ListMonitorServices(ctx context.Context, opts *ListOptions) ([]MonitorService, error) {
 	return getPaginatedResults[MonitorService](ctx, c, "monitor/services", opts)
 }
 
-// ListMonitorServiceByType lists monitor services by a given service_type
-func (c *Client) ListMonitorServiceByType(ctx context.Context, serviceType string, opts *ListOptions) ([]MonitorService, error) {
+// ListMonitorServiceByType gets a monitor service by a given service_type
+func (c *Client) ListMonitorServiceByType(ctx context.Context, serviceType string, _ *ListOptions) (MonitorService, error) {
```
yec-akamai: Does this change mean this endpoint is no longer a list, and only returns a single MonitorService entity?

pmajali: yes it does not return a list now

yec-akamai: Got it, in this case you probably want to build it as other normal GET functions. The header should be something like: 

```suggestion
func (c *Client) GetMonitorServiceByType(ctx context.Context, serviceType string) (MonitorService, error) {
```

test/integration/regions_test.go
```diff
@@ -73,3 +73,26 @@ func TestRegions_kubernetesEnterprise(t *testing.T) {
 	})
 	require.NotZero(t, regionIdx)
 }
+
+func TestRegionsMonitorsSection(t *testing.T) {
+	client, teardown := createTestClient(t, "fixtures/TestRegions_List")
```
yec-akamai: The fixture's name should be unique, otherwise it overwrites the original fixture for another test case: 
```suggestion
	client, teardown := createTestClient(t, "fixtures/TestRegionsMonitorsSection")
```

test/integration/fixtures/TestRegions_List.yaml
yec-akamai, line 1: Once you fixed the fixtures name, can you please revert the change to this file? It's not related to the test case you create

test/integration/fixtures/TestRegions_pgLimits.yaml
yec-akamai, line 1: I think this fixture shouldn't be changed too. Can you please revert it?

test/integration/fixtures/TestRegions_blockStorageEncryption.yaml
yec-akamai, line 1: Same for this one

test/integration/fixtures/TestRegions_kubernetesEnterprise.yaml
yec-akamai, line 1: same here

test/integration/fixtures/TestRegionsAvailability_List.yaml
yec-akamai, line 1: same here

monitor_services.go
```diff
@@ -6,17 +6,32 @@ import (
...
-	return getPaginatedResults[MonitorService](ctx, c, e, opts)
+
+	result, err := doGETRequest[MonitorService](ctx, c, e)
+
+	if result == nil {
+		return MonitorService{}, err
+	}
+
+	return *result, err
 }
```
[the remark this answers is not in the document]
lgarber-akamai: The solution more consistent with our standards would be to make this function return a pointer (`*MonitorService`) returning the results of doGETRequest directly 👍

monitor_dashboards.go
```diff
@@ -51,6 +52,15 @@ type DashboardWidget struct {
 	ChartType         ChartType         `json:"chart_type"`
 	YLabel            string            `json:"y_label"`
 	AggregateFunction AggregateFunction `json:"aggregate_function"`
+	GroupBy           []string          `json:"group_by"`
+	Filters           []DashboardFilter `json:"filters"`
```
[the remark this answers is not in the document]
lgarber-akamai: This change is not necessary because the struct is read-only, slices can be nil, and I don't think the API distinguishes between explicit and implicit nils for this field.

yec-akamai (approved):

PR #825, opened 2025-09-15
Fix `LinodeInterfaceUpdateOptions`

zliang-akamai:
## 📝 Description
1. The `subnet_id` cannot be updated, but it always appears in requests built with `VPCInterfaceCreateOptions`. To address this, create a dedicated VPCInterfaceUpdateOptions for updates.
3. VLAN interfaces cannot be updated at all, so remove them from the update options.


yec-akamai (approved):
Nice work!

PR #827, opened 2025-09-17
[WIP] Add Firewall RuleSet support

komer3:
## 📝 Description

## ✔️ How to Test

## 📷 Preview

### DIFF
--- a/firewall_rules.go
+++ b/firewall_rules.go
@@ -30,7 +30,11 @@ type FirewallRule struct {
 	Ports       string           `json:"ports,omitempty"`
 	Protocol    NetworkProtocol  `json:"protocol"`
 	Addresses   NetworkAddresses `json:"addresses"`
-	Ruleset     *int             `json:"ruleset,omitempty"`
+
+	// Ruleset references a Firewall RuleSet (V2). When provided, this entry
+	// represents a reference to a Rule Set and should be mutually exclusive
+	// with ordinary rule fields.
+	Ruleset *int `json:"ruleset,omitempty"`
 }
 
 // MarshalJSON ensures that when a rule references a Rule Set (Ruleset != nil),
--- a/firewall_rules.go
+++ b/firewall_rules.go
@@ -2,6 +2,7 @@ package linodego
 
 import (
 	"context"
+	"encoding/json"
 )
 
 // NetworkProtocol enum type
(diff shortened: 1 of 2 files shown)


firewall_rulesets.go
```diff
@@ -0,0 +1,137 @@
...
+				return err
+			}
+			boolValue = intValue != 0
+		}
+		if boolValue {
+			raw["is_service_defined"] = json.RawMessage("true")
+		} else {
+			raw["is_service_defined"] = json.RawMessage("false")
+		}
+	}
```
komer3: Note: API seems to be returning 0 or 1 instead of true or false here. Need to verify how to handle this. Right now, during unmarshal, we are translation 0 or 1 to true or false.

firewall_rules.go
```diff
@@ -22,13 +23,51 @@ type NetworkAddresses struct {
...
 	Label       string           `json:"label"`
 	Description string           `json:"description,omitempty"`
 	Ports       string           `json:"ports,omitempty"`
 	Protocol    NetworkProtocol  `json:"protocol"`
 	Addresses   NetworkAddresses `json:"addresses"`
+
+	// FirewallRule references one `Rule Set` by ID. When provided, this entry
+	// represents a reference and should be mutually exclusive with ordinary
+	// rule fields according to the API contract.
+	Ruleset int `json:"ruleset,omitempty"`
```
zliang-akamai:
```suggestion
	RuleSet int `json:"ruleset,omitempty"`
```

Maybe it's `RuleSet` here?

firewall_rulesets.go
```diff
@@ -0,0 +1,151 @@
...
+
+	var intValue int
+	if err := json.Unmarshal(v, &intValue); err != nil {
+		return err
+	}
+
+	raw["is_service_defined"] = json.RawMessage(strconv.FormatBool(intValue != 0))
+
+	return nil
+}
```
zliang-akamai: This API returns both bool and int? 😨

firewall_rules.go
komer3: Ah good catch!

firewall_rulesets.go
komer3: Yeahhh need to confirm this. API spec says it should be a boolean but the actual api seems to be returning 0 or 1

komer3: API doc I was using as ref: https://techdocs.akamai.com/linode-api/reference/post-ruleset

firewall_rulesets.go
komer3: This was a bug. Fix for it has been merged and should be deployed with the next release

firewall_rulesets.go
komer3: We don't need to do this anymore. This was a bug. They are gonna deploy a fix for this in the next release

PR #829, opened 2025-09-24
Fix firewall device for linode interfaces

zliang-akamai:
## 📝 Description

The API doesn't accept `interfaces` key but `linode_interfaces`


lgarber-akamai (approved):
Looks great!

PR #830, opened 2025-09-24
Format everything with `golangci-lint`

zliang-akamai:
## 📝 Description

There is a new rule in the formatter, and we will need to re-format everything to make it be compliant.

Also, disabling `godoclint` linter because we have many comments which are not compliant with it.


jriddle-linode (approved):
LGTM! Thanks for doing this.

PR #914, opened 2026-03-13
Update dependabot configuration to use weekly schedule and group dependencies

zliang-akamai:


zliang-akamai:
succeeded by https://github.com/linode/linodego/pull/936

PR #924, opened 2026-03-31
TPT-4329: Deprecate vpcs in favor of backend_vpcs

dawiddzhafarov:
## 📝 Description

PR deprecates `vpcs` attribute from Create Nodebalancer operation and recommends using `backend_vpcs`, which was added to the request options.

## ✔️ How to Test

```
make test-unit
```

### DIFF
--- a/nodebalancer.go
+++ b/nodebalancer.go
@@ -61,6 +61,13 @@ type NodeBalancerVPCOptions struct {
 	IPv4RangeAutoAssign bool   `json:"ipv4_range_auto_assign,omitempty"`
 }
 
+type NodeBalancerBackendVPCOptions struct {
+	IPv4Range           string `json:"ipv4_range,omitempty"`
+	IPv6Range           string `json:"ipv6_range,omitempty"`
+	SubnetID            int    `json:"subnet_id"`
+	IPv4RangeAutoAssign bool   `json:"ipv4_range_auto_assign,omitempty"`
+}
+
 type NodeBalancerFrontendVPCOptions struct {
 	IPv4Range string `json:"ipv4_range,omitempty"`
 	IPv6Range string `json:"ipv6_range,omitempty"`
@@ -83,7 +90,7 @@ type NodeBalancerCreateOptions struct {
 	// Deprecated: `VPCs` is deprecated in favor of `BackendVPCs`.
 	// This field may be removed in a later release.
 	VPCs         []NodeBalancerVPCOptions         `json:"vpcs,omitempty"`
-	BackendVPCs  []NodeBalancerVPCOptions         `json:"backend_vpcs,omitempty"`
+	BackendVPCs  []NodeBalancerBackendVPCOptions  `json:"backend_vpcs,omitempty"`
 	FrontendVPCs []NodeBalancerFrontendVPCOptions `json:"frontend_vpcs,omitempty"`
 	IPv4         *string                          `json:"ipv4,omitempty"`
 }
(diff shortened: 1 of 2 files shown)


mawilk90 (approved):

test/unit/nodebalancer_test.go
```diff
@@ -201,3 +201,49 @@ func TestNodeBalancer_Create_with_VPCs(t *testing.T) {
...
+
+	var base ClientBaseCase
+	base.SetUp(t)
+	defer base.TearDown(t)
+
+	requestData := linodego.NodeBalancerCreateOptions{
+		Label:  String("Test NodeBalancer VPC"),
+		Region: "us-east",
+		Type:   linodego.NBTypePremium40GB,
+		BackendVPCs: []linodego.NodeBalancerVPCOptions{
```
mawilk90: Shouldn't we add a dedicated `linodego.NodeBalancerBackendVPCOptions` ?

dawiddzhafarov: I mean we could, but it introduces yet another type we would need to maintain (and for users to know), so I decided to keep using already existing type and only modify it.
Pros of adding new type? Maybe more "cleaner" usage (consistent naming).
Cons would be introducing more complexity (new struct, doing exactly the same, just differing in the name).
Let me know what you think.

mawilk90: Yeah, it makes sense. Thanks for explaining this

nodebalancer.go
```diff
@@ -76,13 +76,16 @@ type NodeBalancerCreateOptions struct {
...
-	FrontendVPCs []NodeBalancerFrontendVPCOptions   `json:"frontend_vpcs,omitempty"`
-	IPv4         *string                            `json:"ipv4,omitempty"`
+	Configs    []*NodeBalancerConfigCreateOptions `json:"configs,omitempty"`
+	Tags       []string                           `json:"tags"`
+	FirewallID int                                `json:"firewall_id,omitempty"`
+	Type       NodeBalancerPlanType               `json:"type,omitempty"`
+	// Deprecated: `VPCs` is deprecated in favor of `BackendVPCs`.
+	// This field may be removed in a later release.
+	VPCs         []NodeBalancerVPCOptions         `json:"vpcs,omitempty"`
+	BackendVPCs  []NodeBalancerVPCOptions         `json:"backend_vpcs,omitempty"`
```
zliang-akamai: Should we take this chance to create `NodeBalancerBackendVPCOptions` to avoid confusion?

dawiddzhafarov: I expressed my opinion here: https://github.com/linode/linodego/pull/924#discussion_r3015794313
However if we want to have a separate types for backend and frontend with proper, 1:1 names, that's also fine, probably its better to do it with those changes now rather than later. Let me know if I should do that

test/unit/nodebalancer_test.go
dawiddzhafarov: After rethinking, I pushed a commit that introduces this new type, because it's for a new field introduced to the Request struct. I confused it with modifying existing type (which is not the case here) so it makes more sense, like you and Zhiwei observed.

nodebalancer.go
dawiddzhafarov: https://github.com/linode/linodego/pull/924#discussion_r3026174013

zliang-akamai (approved):

ezilber-akamai (approved):

PR #928, opened 2026-04-06
TPT-3925: Use omitzero for ACLP alerts to support empty lists

yec-akamai:
## 📝 Description

Use omitzero for alerts to make sure it explicitly pass an empty list instead of null. 

Address terraform test issue: https://github.com/linode/terraform-provider-linode/pull/2259

## ✔️ How to Test

```
make test-unit
```


instances.go
```diff
@@ -95,13 +95,13 @@ type InstanceSpec struct {
...
-	TransferQuota int   `json:"transfer_quota"`
-	SystemAlerts  []int `json:"system_alerts,omitempty"`
-	UserAlerts    []int `json:"user_alerts,omitempty"`
+	CPU           int    `json:"cpu"`
+	IO            int    `json:"io"`
+	NetworkIn     int    `json:"network_in"`
+	NetworkOut    int    `json:"network_out"`
+	TransferQuota int    `json:"transfer_quota"`
+	SystemAlerts  *[]int `json:"system_alerts,omitempty"`
+	UserAlerts    *[]int `json:"user_alerts,omitempty"`
```
zliang-akamai: Can we use `omitzero` to avoid breaking change here?

mgwoj (changes requested):
PR title does not match the final changes. Also the changes in the code are not aligned with the changes in the tests.

yec-akamai:
> PR title does not match the final changes. Also the changes in the code are not aligned with the changes in the tests.

Updated the PR title. There is not much to change in the test case since we avoid to make breaking changes. Already verfied the omitzero works as expected across terraform.

mgwoj:
Which tests is showing that this change with `omitzero` works?

zliang-akamai (approved):

mgwoj (changes requested):
Which test is showing that `omitzero` change was needed?

yec-akamai:
> Which test is showing that `omitzero` change was needed?

Added a test case to ensure the marshal result `TestInstanceACLPAlerts_MarshalEmptyList`

mgwoj (approved):

PR #934, opened 2026-04-14
TPT-3909: Integration tests for IAM volumes io_ready

mawilk90:
## 📝 Description

IAM team added `io_ready` field in responses from Volumes endpoints.

NOTE: 
1. PR directly to `main` because the developers code is already there
2. Needs the `iam_beta` tag

## ✔️ How to Test

**Without fixtures:**
`go test -count=1 ./test/integration/ -v -run TestIAM_GetIOReadyFor`

Note: As it runs directly with `go test` you need to have `LINODE_API_VERSION` variable set in your local env

**With fixtures:**
`make test-int TEST_ARGS="-run TestIAM_GetIOReadyFor"`


test/integration/iam_io_ready_test.go
```diff
@@ -0,0 +1,233 @@
...
+	})
+	require.NoErrorf(t, err, "Error creating not attached volume: %v", err)
+
+	volume, err = client.WaitForVolumeStatus(context.Background(), volume.ID, linodego.VolumeActive, 30)
+	require.NoErrorf(t, err, "Error waiting for volume to be active: %v", err)
+
+	volumeList, err := client.ListVolumes(context.Background(), nil)
+	require.NoErrorf(t, err, "Error listing volumes: %v", err)
+	for _, vol := range volumeList {
+		if vol.ID == volume.ID {
```
psnoch-akamai, line 102: How about the case if there is no vol that pass condition? Should it pass?

test/integration/iam_io_ready_test.go
```diff
@@ -0,0 +1,233 @@
...
+		options.Label = label
+	})
+	require.NoErrorf(t, err, "Error creating not attached volume: %v", err)
+
+	volume, err = client.WaitForVolumeStatus(context.Background(), volume.ID, linodego.VolumeActive, 30)
+	require.NoErrorf(t, err, "Error waiting for volume to be active: %v", err)
+
+	volumeList, err := client.ListVolumes(context.Background(), nil)
+	require.NoErrorf(t, err, "Error listing volumes: %v", err)
+	for _, vol := range volumeList {
```
psnoch-akamai, line 101: I would move this loop to function like "findVolumeById" and use returned value for assertions

test/integration/iam_io_ready_test.go
```diff
@@ -0,0 +1,233 @@
...
+	requireSingleAttachedInstanceVolume(t, client, instance)
+
+	volume, err := client.GetVolume(context.Background(), volume.ID)
+	require.NoErrorf(t, err, "Error getting attached volume: %v", err)
+	assertVolumeAttachedToInstance(t, volume, instance)
+
+	err = client.DetachVolume(context.Background(), volume.ID)
+	require.NoErrorf(t, err, "Error detaching volume: %v", err)
+
+	volume, err = client.WaitForVolumeIOReadyStatus(context.Background(), volume.ID, false, 45)
```
psnoch-akamai, line 136: How about assertion for volume value?

test/integration/iam_io_ready_test.go
```diff
@@ -0,0 +1,233 @@
...
+	require.NoErrorf(t, err, "Error waiting for IO Ready status of attached volume: %v", err)
+
+	volumeCloned, err = client.GetVolume(context.Background(), volumeCloned.ID)
+	require.NoErrorf(t, err, "Error getting cloned volume: %v", err)
+	assert.Equal(t, labelCloned, volumeCloned.Label)
+	assert.Equal(t, linodego.VolumeActive, volumeCloned.Status)
+	assert.Equal(t, volume.Region, volumeCloned.Region)
+	assert.Equal(t, volume.Size, volumeCloned.Size)
+	// Cloned volume should not be attached to instance automatically
+	requireSingleAttachedInstanceVolume(t, client, instance)
```
psnoch-akamai, line 208: How about waiting procedure here? How long does it take?

waitfor.go
```diff
@@ -820,3 +820,31 @@ func (client Client) WaitForAlertDefinitionStatus(
...
+		select {
+		case <-ticker.C:
+			volume, err := client.GetVolume(ctx, volumeID)
+			if err != nil {
+				return volume, err
+			}
+
+			complete := (volume.IOReady == status)
+
+			if complete {
```
psnoch-akamai:
```suggestion
			if volume.IOReady == status {
```

psnoch-akamai (changes requested):

test/integration/iam_io_ready_test.go
mawilk90: Yeah, teoretically it may be possible that volumeList does not contain volume.ID and then it will pass with no assertions. I will add `require` to avoid that

test/integration/iam_io_ready_test.go
mawilk90: Is it necessary to extract it to a separate function if it is used only once here? Or you mean to have it for possible future usage?

test/integration/iam_io_ready_test.go
mawilk90: If I got it right, you meant that `client.ListInstanceVolumes` (used inside of `requireSingleAttachedInstanceVolume`) may take longer than expected and fail due to timeout, right?

test/integration/iam_io_ready_test.go
mawilk90: IMO, it is not needed as it is just to wait for volume's io_ready status change (required for the next assertions below). Replaced `volume` with `_` as it is not used anywhere

lgarber-akamai (approved):
Nice work!

test/integration/iam_io_ready_test.go
psnoch-akamai: It would be easier to read and it would be useful in the future

test/integration/iam_io_ready_test.go
psnoch-akamai: I refer to your command line 198. I wonder if we can add step to check status after cloning is successfully done

psnoch-akamai (approved):

test/integration/iam_io_ready_test.go
mawilk90: I check the status of the cloned volume a few lines above using WaitForVolumeStatus

psnoch-akamai (approved):

PR #954, opened 2026-05-06
TPT-4414: Change slice pointer types with `omitzero` to be slice

zliang-akamai:
## 📝 Description

Since https://github.com/linode/linodego/pull/947, we no longer need slice pointer to distinguish nil and initialized empty slice. So re-typing them as non-pointer types.

## ✔️ How to Test

```bash
make test-int
```

```bash
make test-unit
```


ezilber-akamai (approved):
Tests passing locally!

yec-akamai (approved):
Thanks for addressing it! Works well locally

PR #956, opened 2026-05-06
TPT-4113: Add support for Expand Create Linodes Options and Password-less Linodes

ezilber-akamai:
## 📝 Description

Added support for SLADE + CLEO features.

## ✔️ How to Test

`make test-unit`
`make test-int`


test/integration/instances_test.go
```diff
@@ -1149,3 +1150,90 @@ func TestInstance_MaintenancePolicy(t *testing.T) {
...
+		Size:       2000,
+		RootPass:   randPassword(),
+	})
+	require.NoError(t, errCreateDisk)
+	createDiskResponse, err = client.WaitForInstanceDiskStatus(context.Background(), instance.ID, createDiskResponse.ID, linodego.DiskReady, 180)
+	require.NoError(t, err)
+	assert.Equal(t, linodego.DiskReady, createDiskResponse.Status)
+
+	rebuildResponse, errRebuild := client.RebuildInstance(context.Background(), instance.ID, linodego.InstanceRebuildOptions{
+		Image: "linode/alpine3.19",
```
mawilk90: Could you please update the image to linode/alpine3.23 ? linode/alpine3.19 seems to be deprecated. Thanks!

test/integration/instances_test.go
```diff
@@ -1149,3 +1150,90 @@ func TestInstance_MaintenancePolicy(t *testing.T) {
...
+			Type:           "g6-nanode-1",
+			Image:          "linode/debian12",
+			Kernel:         linodego.Pointer("linode/latest-64bit"),
+			BootSize:       linodego.Pointer(8192),
+			AuthorizedKeys: []string{testSSHKeyCreateOpts.SSHKey},
+		},
+	)
+	require.NoError(t, err)
+	assert.Equal(t, "linode/debian12", instance.Image)
+	defer func() {
```
mawilk90, line 1201: Does it make sense to move `defer` function just after `CreateInstance` and before `require` in line 1201? In case of any issue happened during instance creation `defer` will not be registered at all (and possibly instance not deleted). What do you think?

psnoch-akamai: In my opinion it should go just after `require.NoError(t, err)` because there is nothing to remove if this condition not pass

ezilber-akamai: Just pushed up the fix

test/integration/instances_test.go
ezilber-akamai: Just pushed up the fix

lgarber-akamai (approved):

yec-akamai (approved):

mawilk90 (approved):

zliang-akamai (approved):

PR #968, opened 2026-05-13
TPT-4422: Added options structs for POST/PUT endpoints that were missing them

ezilber-akamai:
## 📝 Description

Added dedicated options structs where they were missing.

## ✔️ How to Test

`make test-unit`
`make test-int`

### DIFF
--- a/instances.go
+++ b/instances.go
@@ -380,6 +380,14 @@ type InstanceResizeOptions struct {
 	AllowAutoDiskResize *bool `json:"allow_auto_disk_resize,omitzero"`
 }
 
+type InstanceBootOptions struct {
+	ConfigID int `json:"config_id"`
+}
+
+type InstanceRebootOptions struct {
+	ConfigID int `json:"config_id"`
+}
+
 // InstanceMigrateOptions is an options struct used when migrating an instance
 type InstanceMigrateOptions struct {
 	Type    InstanceMigrationType `json:"type,omitzero"`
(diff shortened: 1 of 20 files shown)


instances.go
```diff
@@ -436,16 +444,18 @@ func (c *Client) DeleteInstance(ctx context.Context, linodeID int) error {
...
+func (c *Client) BootInstance(ctx context.Context, linodeID int, opts InstanceBootOptions) error {
+	var payload InstanceBootOptions
 
-	if configID != 0 {
-		opts = map[string]int{"config_id": configID}
+	if opts.ConfigID != 0 {
+		payload = opts // Use the struct when ConfigID is set
+	} else {
+		payload = InstanceBootOptions{} // Send an empty JSON object when ConfigID is 0
 	}
```
zliang-akamai: Since we already enforce opts, I think we can assume a correct value is in the `opts` and remove this entire `if...else...` block, right?

instances.go
```diff
@@ -462,16 +472,18 @@ func (c *Client) ResetInstancePassword(ctx context.Context, linodeID int, opts I
...
-	opts := make(map[string]int)
+func (c *Client) RebootInstance(ctx context.Context, linodeID int, opts InstanceRebootOptions) error {
+	var payload InstanceRebootOptions
 
-	if configID != 0 {
-		opts = map[string]int{"config_id": configID}
+	if opts.ConfigID != 0 {
+		payload = opts // Use the struct when ConfigID is set
+	} else {
+		payload = InstanceRebootOptions{} // Send an empty JSON object when ConfigID is 0
```
zliang-akamai: Same here, maybe?

instances.go
ezilber-akamai: Yeah good point.

zliang-akamai (approved):

instances.go
```diff
@@ -380,6 +380,14 @@ type InstanceResizeOptions struct {
...
 }
 
+type InstanceBootOptions struct {
+	ConfigID int `json:"config_id"`
+}
+
+type InstanceRebootOptions struct {
+	ConfigID int `json:"config_id"`
+}
+
```
lgarber-akamai, line 390: Just curious, would it make sense to add `omitempty` or `omitzero` to these fields since they're marked as optional in the API docs? 

I'm thinking it might be helpful to do so we can cover the very unlikely scenario that these endpoints have an alternate accepted schema, although that might be overkill.

ezilber-akamai: Yeah I think it makes sense.

lgarber-akamai (approved):
Great work!

PR #978, opened 2026-05-26
TPT-4126: Add support for NB Front-End IP in VPC-MTC + 40Gbps NB-MTC

mawilk90:
## 📝 Description

This PR merges support for NB Front-End IP + 40Gbps features into the main branch for release.

NOTE:
1. Waiting for info from TechDoc if it can be merged/released before their NB changes (planned to be available in Sep)
2. This [PR](https://github.com/linode/terraform-provider-linode/pull/2415) is dependent on it

## ✔️ How to Test

`cd test && make test-unit`
`make test-int TEST_ARGS="-run TestNodeBalancer_Create_With"`


nodebalancer.go
```diff
@@ -66,6 +69,19 @@ type NodeBalancerVPCOptions struct {
 	IPv4RangeAutoAssign bool   `json:"ipv4_range_auto_assign,omitzero"`
 }
 
+type NodeBalancerBackendVPCOptions struct {
+	IPv4Range           string `json:"ipv4_range,omitzero"`
```
ezilber-akamai: Should this be a pointer?

nodebalancer.go
```diff
@@ -66,6 +69,19 @@ type NodeBalancerVPCOptions struct {
...
+	IPv6Range           string `json:"ipv6_range,omitzero"`
```
ezilber-akamai: Same here

nodebalancer.go
```diff
@@ -66,6 +69,19 @@ type NodeBalancerVPCOptions struct {
...
+	SubnetID            int    `json:"subnet_id"`
+	IPv4RangeAutoAssign bool   `json:"ipv4_range_auto_assign,omitzero"`
```
ezilber-akamai: And here

nodebalancer.go
```diff
@@ -66,6 +69,19 @@ type NodeBalancerVPCOptions struct {
...
+}
+
+type NodeBalancerFrontendVPCOptions struct {
+	IPv4Range string `json:"ipv4_range,omitzero"`
```
ezilber-akamai: And here

nodebalancer.go
```diff
@@ -66,6 +69,19 @@ type NodeBalancerVPCOptions struct {
...
+	IPv6Range string `json:"ipv6_range,omitzero"`
```
ezilber-akamai: And here

mawilk90: Hi @ezilber-akamai, all the fields you mentioned above (`IPv4Range`, `IPv6Range`, `IPv4RangeAutoAssign`) seems to be optional and can be skipped in the API request. Does it mean that they need to be a pointers for omitzero anyway? In the current implementation, aren't both empty string ("") and lack of the field treated the same?

zliang-akamai: Yes, the decision was all new omitzero field must be a pointer, and old ones will be migrated to pointer in v3

PR #980, opened 2026-05-27
TPT-4470: Bump major version to v2

zliang-akamai:


lgarber-akamai (approved):
Looks great and I couldn't find any residual v1's, nice work!

ezilber-akamai (approved):

PR #987, opened 2026-06-08
TPT-4275: IP Reservation (IPv4) Project

lgarber-akamai:
## 📝 Description

This pull request merges the changes from the Reserved IPv4 project into the main branch for release.


yec-akamai (approved):

network_reserved_ips.go
```diff
@@ -4,12 +4,40 @@ import (
...
+	Type  string `json:"type"`
+	URL   string `json:"url"`
+}
+
 // ReserveIPOptions represents the options for reserving an IP address
 // NOTE: Reserved IP feature may not currently be available to all users.
 type ReserveIPOptions struct {
-	Region string `json:"region"`
+	Region string   `json:"region"`
+	Tags   []string `json:"tags,omitempty"`
```
zliang-akamai: We need to replace all `omitempty` with `omitzero` since linodego v2

mawilk90 (approved):

network_reserved_ips.go
lgarber-akamai: Good catch! Just pushed up a fix

zliang-akamai (approved):

PR #995, opened 2026-06-15
TPT-4533: Check transport type before setting root certificate and log warning for unsupported transports

zliang-akamai:
`linodego` v2 returns an error from the `NewClient()` when `SetRootCertificate` failed, and setting the certificate requires the transport to be `http.Transport`, but in many cases we are using `oauth2.Transport`, such as Packer Plugin. This will cause the users with `LINODE_CA` environment variable to always see an error returned from `linodego` because `linodego` always auto detect `LINODE_CA` and attempt to set the certificate in `NewClient()`

This PR makes it check if the transport is qualify for custom certificate before automatically attempt to set it.

### DIFF
--- a/client.go
+++ b/client.go
@@ -198,7 +198,8 @@ func NewClient(hc *http.Client) (client Client, err error) {
 	}
 
 	certPath, certPathExists := os.LookupEnv(APIHostCert)
-	if certPathExists {
+
+	if certPathExists { //nolint:nestif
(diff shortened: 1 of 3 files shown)


ezilber-akamai (approved):

yec-akamai (approved):

lgarber-akamai (approved):
LGTM, thanks for the fix!

PR #1019, opened 2026-08-10
feat: [DPS-44447] Add endpoint to expose streams limit per account

odyszews:
## 📝 Description

Adds support for the GET /monitor/streams/quotas endpoint, which exposes stream quota definitions for the authenticated account.

## ✔️ How to Test
```
make test-unit
```


monitor_log_streams.go
```diff
@@ -146,3 +155,8 @@ func (c *Client) DeleteLogStream(ctx context.Context, streamID int) error {
 	e := formatAPIPath("monitor/streams/%d", streamID)
 	return doDELETERequest(ctx, c, e)
 }
+
+// ListLogStreamQuotas returns all quotas for the account.
```
zliang-akamai: Maybe we can make thing clearer by stating what type of quota it is? There are other kinds of quota in Linode API, which is not included in "all quotas" here.

zliang-akamai (approved):

PR #1032, opened 2026-08-27
TPT-4456: Enable integration test for RDMA Linode interface

yec-akamai:
## 📝 Description

Remove skipping the test and allow to run the case. 

## ✔️ How to Test

```
make test-int
```

```
make test-unit
```


test/integration/instance_interfaces_test.go
```diff
@@ -185,14 +188,12 @@ func TestInstance_CreateWithLinodeInterfaces(
...
-	t.Skip("Skipping test because Linode with RDMA interfaces requires manual infra changes at the moment")
-
 	client, fixtureTeardown := createTestClient(t, "fixtures/TestInstance_CreateWithRDMAVPCInterfaces")
 	t.Cleanup(fixtureTeardown)
 
-	// GPUDirect RDMA capability not available for now
+	// GPUDirect RDMA capability not available in all regions, so we hardcode a region that supports it for this test.
 	// testRegion := getRegionsWithCaps(t, client, []linodego.RegionCapability{linodego.CapabilityVPCs, linodego.CapabilityGPUDirectRDMA})[0]
-	testRegion := getRegionsWithCaps(t, client, []linodego.RegionCapability{linodego.CapabilityVPCs})[0]
+	testRegion := "us-rno-1"
```
mawilk90, line 196: Does it make sense to store RDMA region in env vars as you did in Terraform tests?

test/integration/instance_interfaces_test.go
```diff
@@ -240,8 +241,7 @@ func TestInstance_CreateWithRDMAVPCInterfaces(t *testing.T) {
 			opts.RootPass = randPassword()
 			opts.Image = "linode/ubuntu24.04"
 			opts.Region = testRegion
-			// opts.Type = linodego.InstanceRDMAType
-			// opts.HostID = linodego.InstanceRDMAHostID
+			opts.Type = "g3-gpu-rtxpro6000-blackwell-rdma-8"
```
mawilk90, line 244: +1

test/integration/instance_interfaces_test.go
yec-akamai: I think it's not much benefit here in linodego, because we eventually recorded the fixtures and it includes both region and the plan. The major point I use the env vars in terraform is to avoid the test breaks the workflow in the future, when the region is no longer available for us to create rdma.

ezilber-akamai (approved):
Tests pass locally!

zliang-akamai (approved):

PR #1033, opened 2026-08-28
TPT-4603: Added support for NAT Service

ezilber-akamai:
## 📝 Description

Added support and unit tests for NAT Service for VPC.

## ✔️ How to Test

`make test-unit`


vpc_subnet.go
lgarber-akamai, line 136: Do we also need to include NATGateway here?

nat_gateways.go
```diff
@@ -0,0 +1,201 @@
...
+	return doGETRequest[NATGateway](ctx, c, e)
+}
+
+// CreateNATGateway creates a new NAT Gateway
+func (c *Client) CreateNATGateway(ctx context.Context, opts NATGatewayCreateOptions) (*NATGateway, error) {
+	return doPOSTRequest[NATGateway](ctx, c, "networking/natgateways", opts)
+}
+
+// UpdateNATGateway updates the NAT Gateway with the specified id
+func (c *Client) UpdateNATGateway(ctx context.Context, opts NATGatewayUpdateOptions, id int) (*NATGateway, error) {
```
lgarber-akamai, line 146: Looks like the ID and opts are out of order here 👀

vpc_subnet.go
ezilber-akamai: Yep, missed that!

mawilk90 (approved):

PR #1045, opened 2026-09-16
Add NFS spaces and filesystems

komer3:
## Description

Adds Linodego v4beta client support for NFS:

- NFS Space and Filesystem CRUD clients, capacity and usage fields, and lifecycle wait helpers.
- NFS Space and Filesystem Access Policy clients, including VPC, subnet, and Linode ACL models plus mTLS and squash policy options.

These endpoints are needed by the Linode File Storage CSI driver.

Live integration and fixture testing are unavailable because the NFS APIs are not in production. Both commits pass `make build` and temporary HTTP client smoke checks.


komer3:
Could a maintainer apply the `community-contribution` label? This PR does not have the TPT ticket key required by the title check.

nfs_spaces.go
```diff
@@ -0,0 +1,126 @@
...
+	UsedCapacityBytes int64          `json:"used_capacity_bytes"`
+	Created           *time.Time     `json:"-"`
+	Updated           *time.Time     `json:"-"`
+	Tags              []string       `json:"tags"`
+}
+
+// NFSSpaceCreateOptions contains fields accepted when creating an NFS Space.
+type NFSSpaceCreateOptions struct {
+	Label       string    `json:"label"`
+	Description **string  `json:"description,omitzero"`
```
yec-akamai: The double-pointer should not be necessary. Single pointer is good enough. Or is there any specific reason to have double pointer here?

nfs_spaces.go
```diff
@@ -0,0 +1,126 @@
...
+	Tags        *[]string `json:"tags,omitzero"`
+}
+
+// NFSSpaceUpdateOptions contains fields accepted when updating an NFS Space.
+type NFSSpaceUpdateOptions struct {
+	Label       *string   `json:"label,omitzero"`
+	Description **string  `json:"description,omitzero"`
```
yec-akamai: same here^

nfs_filesystems.go
```diff
@@ -0,0 +1,201 @@
...
+		ctx,
+		c,
+		formatAPIPath("nfs/spaces/%d/filesystems/%d", spaceID, filesystemID),
+	)
+}
+
+// GetNFSFilesystemByID returns an NFS Filesystem by its ID. This endpoint requires the v4beta API.
+func (c *Client) GetNFSFilesystemByID(ctx context.Context, filesystemID int) (*NFSFilesystem, error) {
+	return doGETRequest[NFSFilesystem](ctx, c, formatAPIPath("nfs/filesystems/%d", filesystemID))
+}
```
yec-akamai: I wonder if we could give better name to these two functions, because they all fetch by the filesystem id and the current naming is a bit confusing.

Here is my suggestion: 
```
func (c *Client) GetNFSSpaceFilesystem(ctx context.Context, spaceID int, filesystemID int) (*NFSFilesystem, error) 
```
```
func (c *Client) GetNFSFilesystem(ctx context.Context, filesystemID int) (*NFSFilesystem, error)
```
Does it look good to you?

yec-akamai (commented):
Could you please add some unit tests to the new endpoints and functions?

nfs_spaces.go
komer3: The OpenAPI schema allows `description` to be omitted, set to `null`, or set to a string. The distinction matters less during creation, so I’m okay changing this one to `*string`.

nfs_spaces.go
komer3: We need `**string` here. On update, omitting `description` leaves the current value unchanged, `null` clears it, and a string sets it. A single `*string` with `omitzero` cannot distinguish omitted from `null`.

nfs_filesystems.go
komer3: I’d prefer to keep the current names. `GetNFSFilesystem` matches the other nested filesystem operations, while `GetNFSFilesystemByID` clearly identifies the global lookup route. `GetNFSSpaceFilesystem` also sounds like a different resource type. Would you be okay with keeping the current names?

komer3:
Thanks for the review @yec-akamai! I'll work on adding the unit tests using mocks for these funcs

nfs_spaces.go
yec-akamai: Using single pointer *string can support sending explicit null value with omitzero. You don't need double pointer to do so. A non-nil pointer string will not be omitted.

nfs_filesystems.go
yec-akamai: If you can point out which operations match this naming too, we can consider them all together. Still, the `GetNFSFilesystemByID` sounds weird to us. We don't use `byid` in the naming, because it's redundant and the function parameter already tells it's by id. Do you have better idea on this?

yec-akamai: For another one, how about `GetNFSFilesystemInSpace(spaceID, id)`?

komer3: Hmmm thinking about it more, I really like `GetNFSFilesystemInSpace(spaceID, id)`. I'll switch it to that!

nfs_spaces.go
komer3: Maybe I’m missing something, but wouldn’t a single pointer prevent us from explicitly sending `null`? We can set a `*string` to `nil`, but because of omitzero, the field would be omitted from the JSON rather than encoded as `null`. no?

nfs_filesystems.go
komer3: Pushed the change in this commit: https://github.com/linode/linodego/pull/1045/changes/cd360647767c7cba0e7201a5d30493c4b3c9f940

nfs_spaces.go
yec-akamai: Let me better understand the use case: in this case user wants to unset the description, right? How will the API handle it? Is it 
```
description: ""
```
or 
```
description: null
```
are they equivalent? or does api take something else?

yec-akamai: I am asking is because normally API takes `""` to clean up string like description. In this case, singe pointer here is good enough and `""` won't be dropped. However, if this endpoint needs a literal `null`, we should keep it double

yec-akamai:
> Thanks for the review @yec-akamai! I'll work on adding the unit tests using mocks for these funcs

Tested with `make unit-test` and these cases are passing 👍

lgarber-akamai (approved):
LGTM other than Ye's feedback. Thanks for the contribution!

nfs_spaces.go
komer3: testing the API live in devcloud and looks like null is not accepted but the openapi spec says we need that for clearing the description... not sure why there is a mismatch there... 

Lets go with the single pointer. Setting it to empty string should be fine. That works when I tested it

komer3: Pushed the change: https://github.com/linode/linodego/pull/1045/changes/b73f52c4f42f10587fc6223cb5c856a5ca426305

yec-akamai (approved):
Nice work!

